Issue #10, opened 2026-03-17
Preliminary Qrisp Frontend


julfarn:
In order to get some working compilation pipeline going from qrisp to hisep-q as fast as possible, we need to start working on connecting Qrisp to MQT core.

The following approach has been discussed on discord:
 
- Receive textual output from the `jaspr.to_mlir()` function in qrisp.
- Parse to MLIR using the necessary dialects.
  - Work out how to best incorporate the (external) [jasp dialect definition](https://github.com/eclipse-qrisp/Qrisp/tree/main/src/qrisp/jasp/mlir/dialect_definition). Either copy the files, or be inspired by the setup of interaction between [jeff-mlir](https://github.com/PennyLaneAI/jeff-mlir) and mqt core.
- Convert classical dialects like `stablehlo` output by qrisp to builtin MQT-friendly classical dialects using existing tooling.
  - Get inspiration from Catalyst. 
- Write a conversion pass from `jasp` to `QC` (the main effort).

rainij:
@julfarn can we close this issue? We probably already have more specific follow up issues and the base support for the qrisp frontend is already there?

If yes, please close. If not, what is left to be done for this issue?

julfarn:
There are some TODOs in the code, like for instance support for more gates. 

We should open a new issue once demand for more features comes up.

PR #18, opened 2026-03-26
Add jasp -> QC conversion

julfarn:
The second step towards #13.
Currently blocked by [this PR](https://github.com/munich-quantum-toolkit/core/pull/1580) in MQT core and subsequent update of the dependency version on our side.

Added a conversion pass from the `jasp` to `qc` dialects. In addition, introduced a dependency on `stablehlo` so that output from `qrisp` can be handled.

A few todos are still present in the code, but the most urgent features are implemented. 

The remaining steps are now the assembly of a larger pipeline, lowering the classical side to MQT-compatible dialects as well, and thorough testing.


CMakeLists.txt
```diff
@@ -21,10 +21,12 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_BINDIR})
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
 
-# ExternalDependencies MUST be included before SetupMLIR. `mqt-core` calls find_package(MLIR) internally. Setting up
-# MLIR here before they are done with the build would create conflicts.
-include(mlir/cmake/ExternalDependencies.cmake)
+# EarlyExternalDependencies MUST be included before SetupMLIR. `mqt-core` calls find_package(MLIR) internally. Setting
+# up MLIR here before they are done with the build would create conflicts. Other dependencies require SetupMLIR to be
+# finished. These are to be listed in LateExternalDependencies.
```
burgholzer: Let's challenge this at some point. I am pretty sure there is no technical limitation for getting this integration to work without having to be extra careful when to pull in MQT Core.
@denialhaag @simon1hofmann maybe something that either one of you could look into?

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -0,0 +1,397 @@
...
+ */
+struct ConvertJaspCreateQubitsOp final : OpConversionPattern<jasp::CreateQubitsOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(jasp::CreateQubitsOp op, OpAdaptor adaptor,
+                                ConversionPatternRewriter& rewriter) const override {
+    auto loc = op.getLoc();
+    auto extracted = tensor::ExtractOp::create(rewriter, loc, adaptor.getAmount(), ValueRange{});
+    auto index = arith::IndexCastOp::create(rewriter, loc, rewriter.getIndexType(), extracted);
+    auto memrefType = getTypeConverter()->convertType(op.getType(0));
```
burgholzer: Just an idea, but might it be simpler to just directly get the `qc::QubitType` here? It is well-known ahead of time, what the actual type will be here.
Although I also kind of see the charm of the type converter

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -0,0 +1,397 @@
...
+    TRY_CONVERT_GATE("rz", qc::RZOp, 0, 1, 1);
+    TRY_CONVERT_GATE("crz", qc::RZOp, 1, 1, 1);
+    TRY_CONVERT_GATE("s", qc::SOp, 0, 1, 0);
+    TRY_CONVERT_GATE("t", qc::TOp, 0, 1, 0);
+    TRY_CONVERT_GATE("sx", qc::SXOp, 0, 1, 0);
+    TRY_CONVERT_GATE("swap", qc::SWAPOp, 0, 2, 0);
+    TRY_CONVERT_GATE("rxx", qc::RXXOp, 0, 2, 1);
+    TRY_CONVERT_GATE("rzz", qc::RZZOp, 0, 2, 1);
+    TRY_CONVERT_GATE("xxyy", qc::XXPlusYYOp, 0, 2, 2); // TODO: Double check if the angles are ordered correctly
+    TRY_CONVERT_GATE("u3", qc::UOp, 0, 1, 3);
```
burgholzer: Probably worth checking for all of these rotation gates.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+    // TODO: add multi-controlled gates
```
burgholzer: Just pointing out this todo. Is this something you plan to address here or in a follow-up (both is totally fine).

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+    if constexpr (n_controls == 0) {
+      QCOp::create(rewriter, op.getLoc(), operands[targetOperandIndices]...,
+                   extractParam(op, operands[paramOperandIndices + paramIndexOffset], rewriter)...);
+    } else {
+      qc::CtrlOp::create(rewriter, op.getLoc(), operands.take_front(n_controls), [&]() {
+        QCOp::create(rewriter, op.getLoc(), operands[targetOperandIndices + targetIndexOffset]...,
+                     extractParam(op, operands[paramOperandIndices + paramIndexOffset], rewriter)...);
+      });
+    }
```
burgholzer: This is pretty damn cool and compact 🆒 

@denialhaag @simon1hofmann I am pretty sure we could apply the same pattern to almost all conversions we currently have.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -0,0 +1,397 @@
...
+    target.addLegalDialect<QCDialect, memref::MemRefDialect, tensor::TensorDialect, arith::ArithDialect>();
+
+    // Register operation conversion patterns
+    patterns.add<ConvertJaspCreateQuantumKernelOp, ConvertJaspConsumeQuantumKernelOp, ConvertJaspCreateQubitsOp,
+                 ConvertJaspGetQubitOp, ConvertJaspQuantumGateOp, ConvertJaspMeasureOp, ConvertJaspDeleteQubitsOp>(
+        typeConverter, context);
+
+    // Conversion of jasp types in func.func signatures
+    // Note: This currently has limitations with signature changes
+    // TODO: Find solution that works for 1-to-0 conversion of jasp:QuantumStateType
```
burgholzer: Probably something for a follow-up, right?

burgholzer (changes requested):
This is looking really good! I particularly like the code pattern for doing the translation of the gates.

I left a couple of comments/questions inline. But nothing really major that should be blocking this.
In the meantime, the MQT Core ref can also be updated as part of this PR as pointed out by Daniel above 😌 

Overall side comment: it's really time for us to set up some CI here so that we can make sure everything works as expected.
@denialhaag you are probably best equipped for that task.
I am not quite sure yet whether the reusable workflows we use in the MQT could also directly be used here, but chances are non-zero. Even if not, I think it would be worth to try and make that work.

mlir/test/tools/qcc-opt/jasp-to-qc-gates-test.mlir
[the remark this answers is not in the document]
julfarn: No, the overlap in gate set is really good. FYI, I used this list: https://qrisp.eu/reference/Primitives/Gate%20functions.html#gate-application-functions

There's just a few things I have not bothered to implement yet, like multi-controlled gates (possible ofc), daggered gates, and the `parity` operation, which needs to expand to some classical arithmetic. 

With the `memref` support for dynamic lists of qubits, also the translation of the `QubitArray` type is straightforward. Before, I tried to work under the limiting assumption that arrays were of static size, which made the pass much more complicated.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -0,0 +1,18 @@
...
+
+def JaspToQC : Pass<"jasp-to-qc"> {
+  let summary = "Convert jasp dialect to QC dialect.";
+
+  let description = [{
+      This pass converts all operations from the jasp dialect to their equivalent
+      operations in the QC dialect.
+   }];
+
+  // Define dependent dialects
```
```suggestion
```

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -0,0 +1,397 @@
...
+// TODO: add operation support: parity, barrier, slice, fuse, get_size, reset
+
+/**
+ * @brief Converts jasp.create_quantum_kernel by deleting it
+ *
+ * @details
+ * The jasp.QuantumState is used for tracking state in the jasp dialect.
+ * Since QC dialect operations are side-effecting, the initial state creation is dropped.
+ *
+ * Note: For this to work, the IR must be ordered correctly at the start of this pass.
```
rainij:
```suggestion
 * NOTE: For this to work, the IR must be ordered correctly at the start of this pass.
```

Similar to "TODO:" and "FIXME:". Is typically highlighted by IDE if in uppercase.

mlir/test/tools/qcc-opt/jasp-generate-ghz.mlir
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  func.func private @tracerizer() -> tensor<i64> {
+    %c = stablehlo.constant dense<1> : tensor<i64>
+    return %c : tensor<i64>
+  }
+
+}
+
+// CHECK: module
```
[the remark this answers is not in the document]
burgholzer: Fair enough 👍🏼

rainij (changes requested):
Looks very good, just a few small requests for change from my side.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
burgholzer: Yeah, I agree on the readability part. Still better to be explicit though 😌 👍🏼

rainij:
Btw.: Another thing on StableHLO. It depends on LLVM itself. They pin the exact commit in this file https://github.com/openxla/stablehlo/blob/main/build_tools/llvm_version.txt (btw. nice and clean method to document the version). Pinning it is the only way to be really sure that the code builds with llvm as llvm has no stable C++ API.

Now we are in the bad situation were 2 of our dependencies depend on LLVM and MQT does not depend on StableHLO itself, right? So there is potential for a conflict. If I recall correctly catalyst resolved the problem by just pinning to the same version as StableHLO does. But they do not have another LLVM dependency.

I think we need to resolve this problem. Not necessarily in this PR, but we should have a plan on this.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: I added a reminder in a comment at #10

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -0,0 +1,397 @@
...
+ *
+ * @details
+ * The jasp.QuantumState is dropped. The size tensor is extracted to an index
+ * to allocate a memref of qubits.
+ *
+ * Example transformation:
+ * ```mlir
+ * %q_arr, %state1 = jasp.create_qubits %tensor_index, %state0 : !jasp.QuantumState, tensor<i64> -> !jasp.QubitArray,
+ * !jasp.QuantumState
+ * // becomes:
```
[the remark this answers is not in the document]
julfarn: changed

julfarn:
I believe I have adressed all issues. 

@burgholzer, I feel the same about the dependency. Unfortunately, it seems we will also need their passes and not just a dialect, so I see no way around the issue for now, unless we ask qrisp people nicely to get rid of it (they already lower some `stablehlo` to `scf` for technical reasons).

CMakeLists.txt
denialhaag: I have assigned myself to the issue. I'll look into it! 🙂

mlir/cmake/LateExternalDependencies.cmake
```diff
@@ -0,0 +1,13 @@
+include(FetchContent)
+set(FETCH_PACKAGES "")
+
+FetchContent_Declare(
```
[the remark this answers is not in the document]
julfarn: Added a shorter version of this remark and a link to new issue #22

rainij (approved):

mlir/lib/Conversion/JaspToQC/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+  JaspToQCIncGen
+  LINK_LIBS
+  MLIRQCDialect
+  QCCJaspDialect
+  MLIRArithDialect
+  MLIRFuncDialect
+  MLIRTransforms
+  MLIRFuncTransforms)
+
+mqt_mlir_target_use_project_options(QCCJaspoQC)
```
burgholzer:
```suggestion
mqt_mlir_target_use_project_options(QCCJaspToQC)
```

burgholzer (approved):
LGTM except for one small typo that I will quickly apply.

rainij:
> And, as hopefully evidenced by https://github.com/FullStaQD/compiler/pull/18#issuecomment-4171616356, we are willing to act quickly and proactively to work together on this.

Of course we appreciate that, but it is a bad example at the same time. If we would follow my advise and not depend on MQT Core we would not have had the problem in the first place. It was just a one line change after all. @julfarn could have done it with a single commit to the repo instead of reaching out and waiting for a day or two for this to be landed in some other repository.

julfarn:
Hi, thanks for your input!

> From the Qrisp side we would also like to avoid introducing a StableHLO dependency because so far we managed to remain "pure Python".

I don't quite understand. The `jaspr.to_mlir()` function in qrisp emits MLIR containing StableHLO. As far as I can tell, the StableHLO ops originate from proper MLIR (i.e. not xDSL) within the jax framework. So in some sense, you already have this dependency, right? Though I suppose here jax has the responsibility of "shipping the binary" and the internal MLIR functionality is present, but could be hard to access from qrisp's side.

positr0nium:
Thank you for challenging this, @julfarn! I was about to dismiss it but I did a brief check-in with Claude and it indeed found a way to invoke the Jax-shipped binary. The following script removes all StableHLO operations without introducing StableHLO dependency (just Jax). (Also tagging @superlopuh since I think he also was looking for ways to circumvent StableHLO)

```python
"""
Lower a Jaspr quantum program to MLIR with StableHLO ops lowered to
linalg / arith / tensor, while keeping the Jasp dialect ops intact.

Uses only JAX-bundled infrastructure — no external stablehlo or mlir packages.

Pipeline:
  1. Trace a quantum function → Jaspr
  2. Lower Jaspr → JAX MLIR module  (StableHLO + Jasp dialect)
  3. Run  symbol-dce + stablehlo-legalize-to-linalg  on the JAX module
  4. Print to generic MLIR → xDSL → rewrite StableHLO control flow to SCF
  5. Print final MLIR (linalg + arith + tensor + SCF + Jasp)
"""

from jaxlib.mlir import ir, passmanager
from jaxlib.mlir._mlir_libs import _stablehlo, _mlirHlo, _chlo, _jax_mlir_ext

from qrisp import QuantumFloat, h, x, cx, measure, control
from qrisp.jasp.mlir.jaxpr_lowering import lower_jaxpr_to_MLIR
from qrisp.jasp.mlir.jasp_lowering_rules import jasp_lowering_rules
from qrisp.jasp.mlir.quantum_control_flow import fix_quantum_control_flow


def _linalg_mlir_to_xdsl(mlir_text: str):
    """Parse post-linalg generic MLIR into an xDSL module.

    Like ``generic_mlir_to_xdsl`` but also registers linalg, arith, tensor,
    scf, and math dialects that appear after the StableHLO → linalg pass.
    """
    from xdsl.context import Context
    from xdsl.dialects import builtin, func, linalg, arith, tensor, scf
    from xdsl.parser import Parser
    from qrisp.jasp.mlir.xdsl_dialect import JaspDialect

    ctx = Context()
    ctx.allow_unregistered = True
    ctx.load_dialect(builtin.Builtin)
    ctx.load_dialect(func.Func)
    ctx.load_dialect(linalg.Linalg)
    ctx.load_dialect(arith.Arith)
    ctx.load_dialect(tensor.Tensor)
    ctx.load_dialect(scf.Scf)
    ctx.load_dialect(JaspDialect)

    parser = Parser(ctx, mlir_text)
    return parser.parse_module()


def jaspr_to_linalg_mlir(jaspr) -> str:
    """Lower a Jaspr to an MLIR string with StableHLO→linalg applied.

    The pipeline:
    1. Legalize StableHLO arithmetic → linalg  (on the original JAX module)
    2. Then rewrite remaining StableHLO control flow → SCF  (via xDSL)
    """
    # --- Step 1: Lower Jaspr to a JAX MLIR module ---------------------------
    mlir_module = lower_jaxpr_to_MLIR(jaspr, lowering_rules=jasp_lowering_rules)

    # --- Step 2: Register passes and run inside the module's context ---------
    # lower_jaxpr_to_MLIR exits its `with ctx.context:` block, so we must
    # re-enter the MLIR context to use the PassManager.
    ctx = mlir_module.context
    with ctx:
        _stablehlo.register_dialect(ctx)
        _stablehlo.register_stablehlo_passes()
        _mlirHlo.register_mhlo_dialect(ctx)
        _mlirHlo.register_mhlo_passes()
        _chlo.register_dialect(ctx)

        # symbol-dce removes unused private shadow functions that JAX emits.
        # stablehlo-legalize-to-linalg converts arithmetic/data ops to linalg.
        # stablehlo control-flow ops (case, while) are left untouched here.
        pipeline = "builtin.module(" \
                   "symbol-dce," \
                   "stablehlo-convert-to-signless," \
                   "stablehlo-legalize-to-linalg" \
                   ")"

        pm = passmanager.PassManager.parse(pipeline)
        # Disable verifier: stablehlo.case carries !jasp.QuantumState which
        # fails StableHLO type constraints.  The legalizer itself only touches
        # arithmetic ops and leaves case/while alone, so skipping verification
        # is safe here.  The control-flow rewrite to SCF happens next via xDSL.
        pm.enable_verifier(False)
        pm.run(mlir_module.operation)

        # --- Step 3: Print to generic MLIR text ------------------------------
        generic_mlir = mlir_module.operation.get_asm(
            print_generic_op_form=True
        )

    # --- Step 4: xDSL round-trip to rewrite remaining HLO control flow -------
    # stablehlo.case/while carry Jasp quantum types which are fine in SCF
    # but not in StableHLO.  xDSL rewrites them to scf.if/while.
    xdsl_module = _linalg_mlir_to_xdsl(generic_mlir)
    fix_quantum_control_flow(xdsl_module)

    # --- Step 5: Emit the final MLIR text ------------------------------------
    return str(xdsl_module)


# ── Example usage ────────────────────────────────────────────────────────────

def adaptive_bell_state():
    qf = QuantumFloat(3)
    h(qf[0])
    cx(qf[0], qf[1])

    meas_res = measure(qf)

    with control(meas_res == 0):
        x(qf[2])
    
    return measure(qf)


if __name__ == "__main__":
    jaspr = make_jaspr(adaptive_bell_state)()

    print("=" * 72)
    print("  Jaspr → MLIR  (linalg + arith + tensor + Jasp)")
    print("=" * 72)
    print(jaspr_to_linalg_mlir(jaspr))
```

superlopuh:
We've been working on this with the Xanadu folks: https://github.com/xdslproject/xdsl-jax

Am I right in understanding that you already have a custom frontend to StableHLO? If so, it might be worth your time to port the parts of lowering to linalg that you need for your use-case, which would let you drop the dependency.

julfarn:
This is amazing, @positr0nium! Then we can work under the assumption that the IR reaching us has already been converted to `linalg`. 

@superlopuh: Not much of a frontend, we would simply call the stablehlo-to-linalg conversion pass on our side. As soon as we don't need to do that, we can drop the dependency.

positr0nium:
@julfarn yes, you can expect this to become part of Qrisp (possibly a kwarg of ``Jaspr.to_mlir``). Until then maybe just use the provided script for your development.

PR #23, opened 2026-04-02
👷 Add initial CI

denialhaag:
This PR adds an inital CI. The branch is based on the one assocaited to #18.


burgholzer (approved):
Thanks @denialhaag for setting this up. 🙏🏼 
I pushed a couple of commits that
- fix the Windows CI (by adjusting the lit configuration)
- streamline the CI file
- actually enable clang-tidy (the config file was missing)
- fix all warnings reported by clang-tidy

I very much like that this CI setup is so minimal, yet extensive at the same time. The reusable workflows really pay off here.

Once this is in, I will also set up branch protection accordingly so that CI has to pass.

This should give us a pretty good baseline.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+            config: Debug
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-tests-ubuntu.yml@afe75175d32177a7291bbb7bd11b3276f60fc810 # v1.17.15
+    with:
+      runs-on: ${{ matrix.runs-on }}
+      compiler: ${{ matrix.compiler }}
+      config: ${{ matrix.config }}
+      setup-mlir: true
+      llvm-version: 22.1.0
+
+  cpp-tests-macos:
```
rainij: We have to decide at some point what our supported platforms are. I believe we never made a conscious decision here, right? To support some kind of linux is clear as we want to be able to run it in HPC centers. But beyond that nothing is clear.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+            config: Debug
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-tests-windows.yml@afe75175d32177a7291bbb7bd11b3276f60fc810 # v1.17.15
+    with:
+      runs-on: ${{ matrix.runs-on }}
+      compiler: ${{ matrix.compiler }}
+      config: ${{ matrix.config }}
+      setup-mlir: true
+      llvm-version: 22.1.0
+
+  cpp-linter:
```
rainij: What exactly does the linter do? Can you add a comment so that a reader does know what tool is used to do the linting? Similar remark applies to the build tasks.

rainij (changes requested):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-tests-macos.yml@afe75175d32177a7291bbb7bd11b3276f60fc810 # v1.17.15
+    with:
+      runs-on: ${{ matrix.runs-on }}
+      compiler: ${{ matrix.compiler }}
+      config: ${{ matrix.config }}
+      cmake-args: -DMQT_CORE_WITH_GMP=ON
+      setup-mlir: true
+      llvm-version: 22.1.0
+
+  cpp-tests-windows:
```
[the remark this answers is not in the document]
burgholzer: We have been supporting Linux, macOS and Windows on x86 and arm64 for years now over at the MQT. And at least I would have planned to do so here as well.
It is always annoying when packages are only available for a subset of the widely used systems (as shocking as it is, many people in the quantum space still develop on Windows). 
Furthermore, MSVC's debug mode is really good at catching undefined behavior bugs that often go unnoticed on Unix systems. That saved us countless hours of random segfault debugging in the past.
Over the years, I simply set up a Windows VM on my Linux desktop and my MacBook whenever I need to debug something on Windows (which probably only happens a handful of times a year).

I would be in favor of keeping the wide support matrix until we hit friction points.
At that point, I am happy to reconsider.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+  cpp-tests-ubuntu:
+    name: 🇨‌ Test 🐧
+    needs: change-detection
+    if: fromJSON(needs.change-detection.outputs.run-cpp-tests)
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ubuntu-24.04, ubuntu-24.04-arm]
+        compiler: [gcc]
+        config: [Release]
```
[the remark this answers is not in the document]
denialhaag: I have created https://github.com/munich-quantum-toolkit/workflows/issues/352 and https://github.com/munich-quantum-toolkit/core/issues/1622. @rainij, if you have any input on the issues, it's highly appreciated. 🙂

.clang-tidy
rainij: @burgholzer I tried to reproduce locally what happens if I run clang-tidy. I installed version 22 and ran this:

```shell
/usr/local/bin/clang-tidy -p build/my-dev --extra-arg=-std=c++20  mlir/lib/Conversion/JaspToQC/JaspToQC.cpp 
```

It shows me a bunch of warnings like this one (all in the generated files)

```
/workspaces/compiler/build/my-dev/mlir/include/qcc/Dialect/Jasp/IR/JaspTypes.h.inc:52:42: warning: invalid case style for global constant 'dialectName' [readability-identifier-naming]
   52 |   static constexpr ::llvm::StringLiteral dialectName = "jasp";
      |                                          ^~~~~~~~~~~
      |                                          DIALECT_NAME
```

The CI log does not show any output so I am not sure if it does the same. Is this expected, or am I missing a flag to hide the generated files?

rainij (changes requested):

rainij (approved):
Thanks again @denialhaag , especially for the patience with me :).

PR #26, opened 2026-04-07
QIR adaptive lowering

### LINKED ISSUE #25: QIR adaptive profile
In order to support measurement based controlflow we need to be able to lower to LLVM-IR adhering to the QIR [adaptive profile](https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Adaptive_Profile.md).

The goal of this issue is to add a minimal working example in line with our june milestone. Here are some design constraints for the initial version:

- Restrict to latest spec (version 2.0 atm).
- Assume a hardcoded quantum device
  - QDMI support later
  - all-to-all connected device and a small given native gate set (we are mostly interested in the control flow for now).
  - Details will likely gravitate around the needs of HiSEP-Q.
- Support for lowering from `qc` dialect, but make it modular to lower also other dialects in the future.
- Lower standard dialects (to LLVM): `arith`, `cf` primarily.
- Support for the following capabilities:
  - Classical computation with integers
  - IR defined functions and function calls (also: multiple entry_points).
  - Backward branching.
- Make result recording _explicit_ in the IR (either via op, or external function call).
- Restrict to static qubit management (for now) as this is the only one which can be used once mapping to device topology becomes relevant. Dynamic qubit management seems to be at a higher level of abstraction, assuming some later stage which consumes QIR takes care for additional compilation (like routing).

The overall philosophy is that we treat QIR as a low-level format which is meant to be close to the hardware and in particular hardware aware. There are other use cases in which QIR is used as a more high-level exchange format. We concentrate on the former.

Later the QIR sub-pipeline should be able to query device capabilities (e.g. QDMI) in order to choose the right QIR capabilities to compile to. As a simplification we do not explicitly support the base profile right now. It can either be added later for or we use the existing pass from MQT (if it works out of the box).

The first PR does not need to be perfect but it should land fast. We can grow it over time.

rainij:
Resolves #25 .

I also had to synchronize `.clang-format` and `.clang-tidy` as they contradicted each other with respect to how to sort includes. We follow the llvm style now.


mlir/include/qcc/Conversion/ToQIR/CMakeLists.txt
burgholzer, line 1: The naming is a bit inconsistent here. The conversion itself is called `ToQIR`, but then the tablegen target is called `QCCToQIR`.
Could be worth to streamline. I'd be +0.5 for `QCCToQIR` everywhere.

.clangd
burgholzer, line 1: Feels like an unrelated change that would be better suited in a separate PR.

For this case I am not 100% convinced that it is a good idea to disable that check. Although one could argue that this is covered decisively through clang-tidy.

.clang-tidy
burgholzer, line 1: These changes could have been/could be pulled out in a separate PR.

.clang-format
```diff
@@ -1,12 +1,18 @@
...
 IncludeBlocks: Regroup
-IncludeCategories:
-  - Regex: '^".*"'
+IncludeCategories: # consistent with clang-tidy(llvm-include-order)
+  # Our Headers
+  - Regex: '^"qcc/'
     Priority: 1
-  - Regex: '^<.*\.(h|hpp)>'
+  # LLVM Subproject Headers (MLIR)
+  - Regex: '^"mlir/'
```
burgholzer, line 9:
```suggestion
  # LLVM Subproject Headers (MLIR)
  - Regex: '^<mlir/'
```
These are external/system includes. From my past experience people tend to suggest importing these using `<...>` syntax.

.clang-format
```diff
@@ -1,12 +1,18 @@
...
     Priority: 2
-  - Regex: "^<.*>"
+  # LLVM Subproject Headers (LLVM)
+  - Regex: '^"llvm/'
```
burgholzer, line 12:
```suggestion
  - Regex: '^<llvm/'
```
As above

.clang-format
burgholzer, line 1: Feels like an unrelated change that would be better suited in a separate PR.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
burgholzer, line 1: The changes in this file cannot be right and point to a misconfiguration in the changes related to clang-format.

mlir/include/qcc/Dialect/Aux_/IR/AuxDialect.td
```diff
@@ -0,0 +1,30 @@
+#ifndef INCLUDE_QCC_DIALECT_AUX_IR_AUXDIALECT_TD_
+#define INCLUDE_QCC_DIALECT_AUX_IR_AUXDIALECT_TD_
+
+// Copied from https://github.com/eclipse-qrisp/Qrisp/tree/main/src/qrisp/jasp/mlir/dialect_definition
+// TODO: Agree on a single source of truth for the dialect definition.
+// See also issue https://github.com/FullStaQD/compiler/issues/16
+
```
burgholzer:
```suggestion
```
Pretty sure you did not copy this from Qrisp ;-)

mlir/include/qcc/Dialect/Aux_/IR/AuxDialect.td
```diff
@@ -0,0 +1,30 @@
...
+        feature needs some functionality we cannot provide at the moment.
+
+        Please use this dialect only if truly necessary and add a TODO or issue
+        at the respective usage to find a better solution.
+    }];
+
+    let name = "aux";
+    let cppNamespace = "::qcc::aux";
+    let useDefaultTypePrinterParser = 0; // set to 1 once custom types are added.
+}
```
burgholzer, line 24: I am not really a big fan of the generic "Aux" name here; besides the fact that it clashed with Windows (never seen that before 🙃)

But let me pose the bigger question:
Do we really need this dialect? This feels like an easy way out that is way too convenient to take during development of the project as opposed to finding "proper" solutions.
This might not be true here, but from past experience, I can already foresee that this may be used more often than we'd like.

To make a precise proposal: This is only used to add a single operation. What if we were to add that operation to the QC dialect itself? I am still not 100% convinced that we actually need it, but I like that solution more than the notion of a generic auxiliary dialect here.

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
```diff
@@ -0,0 +1,47 @@
...
+
+//===----------------------------------------------------------------------===//
+// Aux operations
+//===----------------------------------------------------------------------===//
+
+class AuxOp<string mnemonic, list<Trait> traits = []> :
+        Op<AuxDialect, mnemonic, traits>;
+
+def RecordBoolOp : AuxOp<"record_bool"> {
+    let summary = "Records a boolean with a symbolic label reference.";
```
burgholzer: I am probably too dense to see it, but where is the "symbolic label reference" coming in here? Shouldn't the label be part of the operation?

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
```diff
@@ -0,0 +1,47 @@
...
+          ...
+          aux.record_bool %m7
+          ...
+        }
+        ```
+
+        To actually translate it to the above QIR runtime function we use a dummy label.
+    }];
+
+    // Change AnyType to FlatSymbolRefAttr
```
burgholzer: Seems to be a leftover comment to me.

```suggestion
```

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
```diff
@@ -0,0 +1,47 @@
...
+    let arguments = (ins I1:$value);
+    let results = (outs);
+
+    // The syntax $label will now correctly handle the @symbol notation
```
burgholzer: Feels like this is referring to something that is no longer there.

mlir/include/qcc/Conversion/ToQIR/Constants.h
```diff
@@ -0,0 +1,62 @@
...
+/// Record a measurement result.
+static constexpr llvm::StringLiteral qirRtResultRecordOutput = "__quantum__rt__result_record_output";
+
+/// Adds a boolean value to the generated output.
+///
+/// Signature: `void(i1,ptr)`.
+///
+/// The second parameter defines a string label for the result value. Depending
+/// on the output schema, the label is included in the output or omitted.
+static constexpr llvm::StringLiteral qirRtBoolRecordOutput = "__quantum__rt__bool_record_output";
```
burgholzer, line 42: From reading https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Adaptive_Profile.md, I do not think that `__quantum__rt__bool_record_output` is actually used in any of the examples. Instead, the examples almost exclusively use `__quantum__rt__result_record_output` to record outputs.
Should we follow that?

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
```diff
@@ -0,0 +1,71 @@
...
+      Essentially lowers the `func` dialect (to llvm) and does a bit of cleanup.
+
+      It assumes that the bodies of the functions are already properly lowered
+      (except `func` related ops).
+   }];
+
+  let dependentDialects = [
+    "mlir::LLVM::LLVMDialect"
+  ];
+}
```
burgholzer: From a logical point of view, this pass should come last in this file.

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
```diff
@@ -0,0 +1,71 @@
...
+      (single bit) measurement result register for each qubit. It refers to the
+      last measurement done at the respective qubit. Hence we use the exact same
+      index for the result of a qubit measurement.
+
+      Assumptions:
+      - All required QIR intrinsics (e.g. `__quantum__rt__initialize`) are already declared in the
+        containing module.
+      - TODO: It expects a global dummy-label to be present for the record intrinsics:
+        `llvm.mlir.global internal constant @".qir_dummy_label"("arbitrary_label\00") {addr_space = 0 : i32}`.
+   }];
```
burgholzer, line 54: The pass description should probably state that a pre-condition is that the `prep-to-qir` pass has been run and a post-condition is that the `finalize-to-qir` pass is run.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+    auto mzFnDecl = moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(qcc::qirQisMZ);
+    auto readFnDecl = moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(qcc::qirRtReadResult);
+
+    if (!mzFnDecl) {
+      return emitMissingDeclError(op, qcc::qirQisMZ);
+    }
+
+    if (!readFnDecl) {
+      return emitMissingDeclError(op, qcc::qirRtReadResult);
+    }
```
burgholzer, line 129:
```suggestion
    auto mzFnDecl = moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(qcc::qirQisMZ);
    if (!mzFnDecl) {
      return emitMissingDeclError(op, qcc::qirQisMZ);
    }

    auto readFnDecl = moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(qcc::qirRtReadResult);
    if (!readFnDecl) {
      return emitMissingDeclError(op, qcc::qirRtReadResult);
    }
```
Probably just personal preference, but I think this is slightly cleaner.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    // TODO: This holds only for HiSEP-Q.
+    // NOTE: Qubit and result pointer share the same index.
+    auto qubit = op.getQubit();
+    auto ptrs = qubitsToPtrs(rewriter, {qubit, qubit});
+
+    LLVM::CallOp::create(rewriter, op.getLoc(), mzFnDecl, ptrs);
+    auto callReadOp = LLVM::CallOp::create(rewriter, op.getLoc(), readFnDecl, {ptrs[1]});
```
burgholzer: Given this assumption, why create the constants and casts twice?

```suggestion
    // TODO: This holds only for HiSEP-Q.
    // NOTE: Qubit and result pointer share the same index.
    auto qubit = op.getQubit();
    auto ptrs = qubitsToPtrs(rewriter, {qubit});

    LLVM::CallOp::create(rewriter, op.getLoc(), mzFnDecl, {ptrs[0], ptrs[0]});
    auto callReadOp = LLVM::CallOp::create(rewriter, op.getLoc(), readFnDecl, {ptrs[0]});
```
Might be worth to add a `qubitToPtr` function that avoids materializing a SmallVector.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+struct UnitaryLowering : public ConversionPattern {
+  UnitaryLowering(TypeConverter& converter, MLIRContext* ctx)
+      : ConversionPattern(converter, MatchAnyOpTypeTag(), 1, ctx) {}
+
+  LogicalResult matchAndRewrite(Operation* op, ArrayRef<Value> /*operands*/,
+                                ConversionPatternRewriter& rewriter) const override {
+    auto unitaryOp = dyn_cast<qc::UnitaryOpInterface>(op);
+    if (!unitaryOp || !isa<qc::QCDialect>(op->getDialect())) {
+      return failure();
+    }
```
burgholzer, line 172: Probably stupid me, but couldn't you directly match for `qc::UnitaryOpInterface` in the first place?

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+  LogicalResult insertRtInit() {
+    func::FuncOp funcOp = getOperation();
+    auto moduleOp = funcOp->getParentOfType<ModuleOp>();
+    auto* ctx = funcOp.getContext();
+
+    auto initFnDecl = moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(qcc::qirRtInit);
+
+    if (!initFnDecl) {
+      return moduleOp.emitError() << "missing required declaration of QIR runtime function: " << qcc::qirRtInit;
+    }
```
burgholzer, line 249: Why not use `emitMissingDeclError` here?

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+    getOperation()->walk([&](Operation* op) {
+      if (isa<qc::StaticOp>(op)) {
+        toErase.push_back(op);
+      }
+    });
+
+    for (Operation* op : toErase) {
+      op->erase();
+    }
+  }
```
burgholzer, line 297: Wouldn't this be more efficient as a pattern that only matches StaticOp and removes it?

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
```diff
@@ -0,0 +1,151 @@
...
+      // LLVM seems to normalize to i32 values. Even if we would use e.g. a i1
+      // attribute for the boolean flags `mlir-translate` would still force it
+      // to i32. Hence we stick to i32 always here.
+      return LLVM::ModuleFlagAttr::get(ctx, behavior, builder.getStringAttr(name), builder.getI32IntegerAttr(val));
+    };
+
+    // NOTE: Missing flags are implicitly set to "false".
+    const SmallVector<Attribute> flags = {
+        createFlag(LLVM::ModFlagBehavior::Error, "qir_major_version", 2),
+        createFlag(LLVM::ModFlagBehavior::Max, "qir_minor_version", 0),
```
burgholzer: May as well bump to 2.1 here already since we do not make use of dynamic memory management.

```suggestion
        createFlag(LLVM::ModFlagBehavior::Max, "qir_minor_version", 1),
```

mlir/lib/Conversion/ToQIR/FinalizeToQIR.cpp
burgholzer, line 1: This part feels like it could have been integrated into the prior parts already.
The qcc entrypoint attribute can be replaced when the QIR attributes are added. (they are added to the same function).
The remaining LLVM pass can run as part of the pipeline that is already set up of the conversion.
But maybe I am missing something here.

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
burgholzer, line 1: Is the split into preparation and conversion here really that helpful compared to combining this in one dedicated ToQIR conversion pass? It just feels like most of the code here could equally reside in the main conversion code and would still fulfill its purpose.
Although I generally like modularity and splitting up large files, I am not sure this is warranted here.

mlir/test/tools/qcc/example_ghz.mlir
```diff
@@ -0,0 +1,52 @@
+// RUN: qcc -o - %s | FileCheck %s
+
+/// Prepare GHZ state (without loop so far).
+///
+/// TODO: This test only tracks what can be done so far with `qcc`. If more functionality
+/// evolves this test can go and be replaced by a better test.
```
burgholzer: This test feels slightly incomplete as it does not _use_ `%m0` to `%m2`.
They could/should at least be recorded as outputs, if we are adding that operation as part of this PR.

mlir/test/tools/qcc-opt/ToQIR/export-to-qir-full-pipeline.mlir
```diff
@@ -0,0 +1,77 @@
...
+    %m0 = qc.measure %0 : !qc.qubit -> i1
+
+    cf.cond_br %m0, ^bb_true, ^bb_exit
+    ^bb_true:
+      qc.x %0 : !qc.qubit
+      cf.br ^bb_exit
+    ^bb_exit:
+
+    %m1 = qc.measure %0 : !qc.qubit -> i1
+    aux.record_bool %m1
```
burgholzer, line 18: This is related to my general comment on the new Aux Operation for recording outputs. After having reviewed the entire PR in quite a bit of detail, I am beginning to more and more see the use for such an operation as it acts as a consumer of the measurement output and does not leave the measurement output as an unused value. Kind of feels like it makes the program more "complete" in a sense.

burgholzer (changes requested):
Thanks @rainij for pushing this. I looked through this in quite a bit of detail now. You'll find a couple of comments with suggestions as well as points for discussion inline.
I hope most of these make sense. My brain was a little slow today, so please excuse if some of the questions/points turn out to be stupid. I just wanted to make sure to get feedback back to you as soon as possible.

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -12,10 +16,18 @@ void buildQuantumPipeline(mlir::PassManager& pm) {
...
 
   // QC → QIR Conversion
-  pm.addPass(mlir::createQCToQIR());
+  pm.addPass(qcc::createPrepToQIR());
+
+  mlir::OpPassManager& fpm = pm.nest<mlir::func::FuncOp>();
+  fpm.addPass(mlir::createArithToLLVMConversionPass());
+  fpm.addPass(qcc::createConvertQCToQIR());
+
+  pm.addPass(mlir::createConvertControlFlowToLLVMPass());
```
julfarn, line 82: The output of the qrisp import pass has `scf`, not `cf`. So at some point we also need to make that conversion.

mlir/include/qcc/Conversion/ToQIR/Constants.h
```diff
@@ -0,0 +1,62 @@
...
+/// Convert a measurement result to a bool.
+///
+/// Signature `i1(ptr readonly)`.
+static constexpr llvm::StringLiteral qirRtReadResult = "__quantum__rt__read_result";
+
+// TODO: it is unclear how our compiler should handle it.
+/// Record a measurement result.
+static constexpr llvm::StringLiteral qirRtResultRecordOutput = "__quantum__rt__result_record_output";
+
+/// Adds a boolean value to the generated output.
```
RiccardoRomanello, line 36: This is mostly a doubt rather than an actual suggestion. Any reason why using the `adds` instead of just imperative `add`? Because I saw in some places one version, in some places the other. Is there a rationale behind it? (Also for future coding purposes)

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
```diff
@@ -0,0 +1,71 @@
+include "mlir/Pass/PassBase.td"
+
+def PrepToQIR : Pass<"prep-to-qir", "mlir::ModuleOp"> {
+  let summary = "Very first pass to apply if you want to convert to QIR (adaptive).";
```
RiccardoRomanello: I would change the summary to something like `Very first pass to apply for a conversion to QIR (adaptive).` Just to avoid the `you want to` which sounds a bit informal to me. What do you think?

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
```diff
@@ -0,0 +1,71 @@
...
+}
+
+def ConvertQCToQIR : Pass<"convert-qc-to-qir", "mlir::func::FuncOp"> {
+  let summary = "Lower the QC dialect into QIR (adaptive) conforming LLVM dialect.";
+
+  let description = [{
+      The pass anchors to a function and guarantees the following:
+
+      - The QIR initialization function is inserted at the start of the function.
+      - Required QIR attributes (like `required_num_qubits`) are added to the function. Via `passthrough`.
```
RiccardoRomanello, line 29: An extra `.`?

mlir/include/qcc/Conversion/ToQIR/CMakeLists.txt
rainij: My feeling is that it is actually more consistent if I leave it as is. If I compare with Jasp: The folder is called `JaspToQC` and the target is called `QCCJaspToQC`. We prefix our targets always with `QCC` but most of the other things (like folders) not as it would be redundant (the top-level include dir is already `qcc`).

But I believe I understand where the confusion comes from: `ToQIR` does not state were it starts from. This is intentional: It is not just for one single pass it should contain all passes related to QIR lowering. This includes `qc` to QIR but also the support passes like prep and finalize. In the future also other dialect conversions. I would like to make it modular. This would also be consistent with other builtin passes like `-arith-to-llvm` which focusses on just one dialect.

.clangd
rainij: I agree, it is really not necessary to do it here.

rainij: I remove the file

mlir/include/qcc/Conversion/ToQIR/Constants.h
rainij: Good point. I think we do not have a guideline here.

mlir/include/qcc/Conversion/ToQIR/Constants.h
rainij: I am aware of that. But the bool version actually fits nicer into what is already there. I still need this aux dialect but it is the smallest change I could come up with. 

It would be possible to use the result_record function: We could naturally invent an op on our side which says "record whatever was last measured for qubit i" (so it would take a qubit as input). Makes total sense. But I do not know if HiSEP-Q would support it so I went with the (IMO) more natural solution involving the bool. 

In any case we will definitely improve this in the future. It is really just a workaround for now to not block the PR. IMO the workaround isn't even too bad.

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
rainij: DONE.

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
rainij: DONE.

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
rainij: I wanted to make it explicit what is really needed. Not everything the prep pass does is truly necessary for this pass to do its job. But I added a sentence which recommends the usage of these passes. This makes total sense.

mlir/include/qcc/Dialect/Aux_/IR/AuxDialect.td
rainij: Thanks for spotting this. I actually noticed it myself and removed it in one of my cleanup commits. Not sure why it wasn't removed.

mlir/include/qcc/Dialect/Aux_/IR/AuxDialect.td
rainij: It is really not meant as a final solution. And it is also not meant to advocate laziness. I am totally in favor of adding something like this to QC dialect itself or any of the dialects we plan right now for QCC. But I think this thing alone would deserve a whole PR, a discussion around it, etc. Hence I think it is justified to add a preliminary solution to this PR to not block it. It also serves as a way to field test the idea.

The Aux dialect and its weird name are really intended as a warning. I would really like to leave it here as a way to quickly move forward. IMO it is a really nice solution to isolate workarounds.

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
rainij: This is an error on my side. It refers to an older version of this op which refers to a label by symbol explicitly. Since we have no support for such labels at the current IR I figured I could as well leave it out here and make the op cleaner. I forgot to update the summary then.

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
rainij: Thanks, traces of AI usage :D .

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
rainij: Correct. Removed.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij: Yeah it is not really misconfigured, it is just about `<foo>` vs `"foo"`. It related to this CI vs prek vs tidy issue (the tools fighted each other). Let us discuss in another thread.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: Totally fine with that. Changed it accordingly. (I had to do it manually as I already have changes in my working tree)

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+using namespace mlir;
+using namespace qcc;
+
+/// Map any of the (unitary) gate-ops to their QIR QIS function declaration if possible.
+///
+/// Returns an empty string upon failure. Succeeds iff the gate is among the supported native gate set.
+///
+/// TODO: The native gate set is currently hardcoded.
+static StringRef mapUnitaryToQIS(qc::UnitaryOpInterface unitaryOp) {
```
RiccardoRomanello, line 37: Why not using a `FailureOr<StringRef>` return type? Would stress more the wrong paths that can happen during the conversion

mlir/include/qcc/Conversion/ToQIR/CMakeLists.txt
burgholzer: I can get behind that reasoning 👍🏼 makes sense!

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+        .Case<qc::HOp>([](auto) { return qcc::qirQisH; })
+        .Default([](auto) { return ""; });
+  }
+
+  if (unitaryOp.getNumControls() == 1) {
+    auto ctrlOp = cast<qc::CtrlOp>(unitaryOp);
+    auto bodyOp = ctrlOp.getBodyUnitary();
+
+    return llvm::TypeSwitch<Operation*, StringRef>(bodyOp)
+        .Case<qc::XOp>([](auto) { return qcc::qirQisCX; })
```
RiccardoRomanello, line 50: Shouldn't it be a `.Case<qc::CXOp>` in the pattern?

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: Good idea. Added the `qubitToPtr` method and also use it where appropriate (including the implementation of the `qubitsToPtrs` function). 

I still create the constant and the inttoptr cast twice as it is much clearer to me what is going on (and generalizes). Actually with the `qubitToPtr` function it became even clearer. The CSE pass can efficiently remove the duplicates (it is needed anyway somewhere after this pass).

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: I remember that I wanted this. There was some problem with that. I believe the _conversion_ variant of the rewrite pattern does not support matching against interfaces. It was something surprising.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+      return failure();
+    }
+
+    auto moduleOp = op->getParentOfType<ModuleOp>();
+
+    auto qisName = mapUnitaryToQIS(unitaryOp);
+    auto fnDecl = moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(qisName);
+    if (!fnDecl) {
+      return emitMissingDeclError(unitaryOp, qisName);
+    }
```
RiccardoRomanello, line 180: Would remove the space between row 168 and 170, since `moduleOp` is only used in the scope of the next 3 lines... They 'belong' together

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: It slightly different (QIS vs runtime functions). I renamed the function to make it clearer: `emitMissingQISDeclError`.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    return llvm::success();
+  }
+
+  uint64_t getRequiredNumQubits() {
+    const func::FuncOp funcOp = getOperation();
+    uint64_t numQubits = 0;
+    funcOp->walk([&](qc::StaticOp op) -> void {
+      auto index = op.getIndex();
+      numQubits = std::max(numQubits, index + 1);
```
RiccardoRomanello, line 267: Could this potentially be unsafe for multiple functions? For the matter of parallelization and race conditions and so on and so forth.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: No actually it is better with a walk. Also: it has to happen as the final cleanup (not with the other patterns together) because only then I can guarantee that they are unused. My gut feeling is also that a walk is more efficient performance wise. In any case, it is shorter to implement it this way (in terms of lines of code).

But good that you mentioned it. This small vector allocation is a bit silly. I can delete the stuff right away making the function essentially a one liner. Not sure why I did this in the first place. Simplified the implementation.

RiccardoRomanello (changes requested):

mlir/lib/Conversion/ToQIR/FinalizeToQIR.cpp
rainij: I can understand this. I would like to keep it for now as I feel not 100% save deleting the `qcc.entry_point` already in the other pass. I wanted to have the option that some other pass can still rely on the flag being there (without needing to look it up in some passthrough stuff). 

It is a nice option to have the pass for now, but if it truly proves to be redundant I would consider removing it in a different PR. Does this work for you?

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
rainij: Here I am really strongly in favor of having prep and qc separated. I see the qc pass as a sibling of the arith-to-llvm pass. Both passes can work at the function level which makes conceptually total sense as arith and qc contain only local ops. That is a really nice feature. 

If we ever have very large QIR modules with many functions this would even matter due to parallization (the prep stuff has to happen at the module level).

Also: It is likely we want to export other dialects to llvm. This way we can reuse the prep pass.

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
rainij: OK.

mlir/test/tools/qcc-opt/ToQIR/export-to-qir-full-pipeline.mlir
rainij: Yeah this is one use case. As I said this here is just a POC for this feature. I would definitely like to leave it in this PR, but not as a lasting solution. Instead we should have a discussion on this whole output recording topic (discord, issues, etc) and implement a satisfying solution. I guess this takes some time - but we have it. Eventually this `aux` thingy can go away (actually "must" IMO).

mlir/test/tools/qcc/example_ghz.mlir
rainij: I agree. Added the record ops.

.clang-format
rainij: In principle I would agree but the tools were truly fighting each other: prek/clang-format vs clang-tidy. I have setup IDE integration for format and tidy and the rule contradicted. Hence prek stopped me from commiting. Even if I deactivate locally the pre-commit hook in the CI would fight the tidy check.

So let us just consider it a quickfix. I will open a PR on this after my lunch break.

.clang-format
rainij: Lets discuss in the dedicated PR. I am inconclusive on that actually.

.clang-format
rainij: Lets discuss in the dedicated PR. I am inconclusive on that actually.

.clang-tidy
rainij: As said in the other thread, it was an emergency fix to shut up fighting tools. I really had to do this to move on. I will open the PR where we can make a conscious decision.

rainij:
@burgholzer @EnriqueNB I went over the first round of comments. I still have to merge main, but I will do this after my break. But you can already look over my replies if you like.

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
burgholzer: Makes sense to me and the added sentence fits nicely 👍🏼

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
```diff
@@ -0,0 +1,74 @@
...
+      Essentially lowers the `func` dialect (to llvm) and does a bit of cleanup.
+
+      It assumes that the bodies of the functions are already properly lowered
+      (except `func` related ops).
+   }];
+
+  let dependentDialects = [
+    "mlir::LLVM::LLVMDialect"
+  ];
+}
```
burgholzer: Strange that this was not fixed by pre-commit, but this file seems to be missing a newline at the end.

mlir/include/qcc/Dialect/Aux_/IR/AuxDialect.td
burgholzer: I can get behind the motivation of isolating workarounds.
And I also do not want to block this PR by any means. So let's go with Aux for now.
We should monitor very closely over the future whether we actually resolve workarounds at some point or whether we are creating a Frankenstein monster.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
burgholzer: Ah, surprising indeed. Might be something to revisit eventually in the future.
Feel free to resolve here though.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
burgholzer: Couldn't you still reuse the same function by simply having the string say "missing required declaration of QIR function: "?
It's not too important anyway, as it's just a one liner but it caught my eye anyway.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
burgholzer: The new version seems much cleaner 👍🏼 feel free to resolve.

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
burgholzer: Fair points! I did not think so far ahead. Thanks for the explanation!

mlir/test/tools/qcc-opt/ToQIR/export-to-qir-full-pipeline.mlir
burgholzer: > Eventually this aux thingy can go away (actually "must" IMO).

Definitely agree on that.

Let's leave it in for now and start a separate discussion thread for what we want to do about this.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
burgholzer: In the previous configuration that we had, the system includes where separated from the llvm/mlir includes (even when both used the `<...>` syntax). So I am pretty sure there is a solution here. But that's nothing to discuss here. Let's move this to a separate discussion.

mlir/include/qcc/Conversion/ToQIR/Constants.h
burgholzer: Fine as a workaround for now.

We shouldn't tailor too many of our developments to HiSEP-Q in particular as opposed to the broader community. If the broader QIR community is using the Result recording version, I believe it is warranted to follow that and rather urge the HiSEP-Q conversion or the target itself to adapt.
Something for a follow-up though. Feel free to resolve here.

mlir/lib/Conversion/ToQIR/FinalizeToQIR.cpp
burgholzer: Yeah. Fine with me.

burgholzer (commented):
I think I commented on all the open review comments. Let me know if I missed something.

mlir/test/tools/qcc/example_ghz.mlir
julfarn: Would be really natural to return the values here. This is not strictly allowed by QIR, but in the kernel/classical code interfacing we will need to find realistic solutions within the project. I could see us deviating from the output recording standard, as this feels a bit awkward for "smarter" QCS's.

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
rainij: Maybe I know where this comes from: I had to turn off prek (temporarily) when committing changes for this file as prek kept deleting the content again and again. I added the newline now (will soon be pushed).

mlir/include/qcc/Conversion/ToQIR/Constants.h
rainij: I brought it up on discord so I close it.

mlir/include/qcc/Conversion/ToQIR/Constants.h
rainij: Yes I agree. I am in favor of generality too.

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
rainij: I do not understand. You can also use the code suggestion feature to point out what you mean ("Add a suggestion" button).

mlir/lib/Compiler/Pipeline.cpp
rainij: I rather certain the is a dedicated pass for the scf to cf conversion. This can be added easily. I probably have to do it soon anyway as your branch is already merged and I have to merge in your stuff.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij: Yeah I know what you mean, one can formulate the regex this general. I take this into account in the upcoming PR.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: Yeah I know `FailureOr` and I am glad it exists, but I tend to not use it for simple things like this. 

The problem with this thing is that there is no native support for these result types / optional types etc in C++. In rust it would be a different story. In C++ the code to pack/unpack and check the stuff is so much boilerplate (for simple examples like this) that is spoils readability significantly. My feeling is that in C and C++ it is a perfectly valid pattern to use `""` as a marker for a missing value. Similar in spirit to `nullptr`.

I later use it like so

```cpp
auto qisName = mapUnitaryToQIS(unitaryOp);
    auto fnDecl = moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(qisName);
    if (!fnDecl) { /* ERROR */ }
```

and avoid double checking the stuff as I can leverage that a symbol must not have an empty name. This way I emulate the error bubbling of rust a little bit (well I still have this final if statement but better than nothing).

Also the implementation of the function itself looks nicer (less boilerplate). I might be wrong but I believe it is a valid C++ design pattern inspired by its closeness to assembly language.

I would actually be interested in the opinion of @burgholzer .

Btw @julfarn would probably celebrate the superiority of rust if he was here ;).

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: No this one does not exist. Controlled gates are provided via the modifier op `CtrlOp` applied to an `XOp`. For that reason I have to top-level if-statements in the function.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: Makes sense. Done. I also renamed it yet again to `emitMissingQIRDeclError` to meet this generality in the name.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij: No I see no race condition here. The number of qubits is per function anyway. The pass anchors to a function so if there are multiple instances of the pass they would operate on different functions.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
RiccardoRomanello: Ok I get it, I lost the fact that you were checking on the body unitary, not the whole operation! My bad

rainij:
@burgholzer @julfarn I finally managed to merge the main branch. This was actually quite some work. Unfortunately the test `jasp-generate-ghz.mlir` could not be preserved. The frontend made by Julian and the QIR conversion made by me work under different assumptions and do not meet in the middle. 

The changes required to make both parts work together require extensive work. I would ask @RiccardoRomanello to work on this in one or multiple upcoming PRs. He, Julian and me should have a meeting together to figure out the details.

The good news is that my test for `qcc` the `example_ghz.mlir` still works. The frontend just washes over it doing nothing and my backend can take it up. The follow up PR could in principle try to reformulate this test (which is simpler in structure than Julians test) to qrisp hence establishing a full work through from qrisp to QIR.

@burgholzer any objections?

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
burgholzer: Hm. I can see both sides here.
It is hard to draw the line when the complexity is high enough to warrant using `FailureOr`.
I am not particularly opposed to using an empty string as a kind of failure mode. The only thing that bugs me a bit (and I am surprised clang-tidy is not complaining; but this is probably due to LLVM's `StringRef`) is that
```
if (!fnDecl)
```
is an implicit check and would explicitly be written as
```
if (!fnDecl.empty()) // or if (fnDecl != "")
```
At which point it is no longer much prettier than the `FailureOr` code.

But, to be fair, I really don't have a strong opinion here. The above is just what came to my mind.

burgholzer (approved):
Nice work @rainij 👏🏼 
This looks pretty good now. No objections from my side on getting this in and working on the backend-frontend connection in follow-up PRs.

mlir/include/qcc/Conversion/ToQIR/Constants.h
EnriqueNB: Yes, there is a rationale and we should be consistent about it. You use the third person for descriptive purposes: describing functions, classes, modules; explaining behavior, constraints; writing API docs. The imperative for prescriptive purposes –telling the user what to do–: inline comments that guide usage, error messages, CLI help text, tutorials. 

For example, here is Google style guide: https://developers.google.com/style/reference-verbs. We should be consistent with this from the beginning because if not the documentation becomes a mess and difficult  to change. 

In this case, the comment is describing a function so we use the third person.

mlir/include/qcc/Conversion/ToQIR/Constants.h
```diff
@@ -0,0 +1,62 @@
...
+
+/// Initializes the execution environment.
+///
+/// Signature: `void(ptr)`.
+///
+/// Sets all qubits to a zero-state if they are not dynamically managed. Must be
+/// called right at the start of an entry-point.
+static constexpr llvm::StringLiteral qirRtInit = "__quantum__rt__initialize";
+
+/// Convert a measurement result to a bool.
```
EnriqueNB: For example, here it would be _Converts_ because it is describing a function. @rainij can you make sure that the usage of the third person is consistent? There are many cases in the code.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
RiccardoRomanello: If you think in this case an empty string is ok to signal the problem, then I am pretty fine with that!
I am more of a fan of the `-1` return value when dealing with integers than to empty strings. 

Maybe in the future when the problems become 'bigger' we could start thinking about a standard way of dealing with this to be all on the same line

RiccardoRomanello (approved):
I went through both the comments and the code and everything seems fine to me!

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
```diff
@@ -0,0 +1,74 @@
...
+
+      TODO: We hardcode hardware capabilities for now as we have no way to query
+      those. It is currently tailored towards HiSEP-Q. In particular we use
+      static qubit and result management. Note that HiSEP-Q has a dedicated
+      (single bit) measurement result register for each qubit. It refers to the
+      last measurement done at the respective qubit. Hence we use the exact same
+      index for the result of a qubit measurement.
+
+      Assumptions:
+      - All required QIR intrinsics (e.g. `__quantum__rt__initialize`) are already declared in the
```
EnriqueNB: Why do you call it a QIR intrinsics? A consider an intrinsic to be a LLVM intrinsic function. There is no concept of intrinsic in QIR. The is a QIR runtime function call.

EnriqueNB (approved):
Looks good to me. There are some workarounds but this could be handled in other merge requests. I approve, please see the comments and take them into account if you have time.

mlir/include/qcc/Conversion/ToQIR/Constants.h
rainij: OK I changed a few of the docstrings, I cannot guarantee to find them all. We can be more careful next time if we agree on a common guideline.

mlir/include/qcc/Conversion/ToQIR/ToQIR.td
rainij: OK, I changed it to the very neutral "function". The term "runtime function" also does not feel right.

PR #31, opened 2026-04-13
Qrisp frontend: Assembly of classical pipeline.

julfarn:
This completes, for now, the qrisp integration according to #10 from the FullStaQD side. 

 - Removed `stablehlo` dependency, as we can rely on qrisp to lower to `linalg` on the classical side. 
 - Added conversion of Rank-0-Tensors like `tensor<i64>` to plain values like `i64` in `JaspToQC` pass.
 - Removed appearance of `!jasp.QuantumState` type that remained in function signatures after `JaspToQC` pass.
 - Assembled pipeline to lower classical operations to `memref`, `arith`, and `scf` dialects.


mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -1,22 +1,51 @@
...
+  // Initial cleanup
+  pm.addNestedPass<mlir::func::FuncOp>(mlir::createCanonicalizerPass());
+  pm.addNestedPass<mlir::func::FuncOp>(mlir::createCSEPass());
+
+  // Bufferization prep
+  pm.addPass(mlir::bufferization::createEmptyTensorToAllocTensorPass());
+
+  mlir::LinalgDetensorizePassOptions detensorizeOptions;
+  detensorizeOptions.aggressiveMode = true;
+  pm.addNestedPass<mlir::func::FuncOp>(mlir::createLinalgDetensorizePass(detensorizeOptions));
```
rainij, line 39: Could you add a one-line comment roughly explaining it purpose? My feeling is that it is responsible for 0-D tensors and must happen before teh one shot bufferize (which directly turns it into memory allocations otherwise).

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -1,22 +1,51 @@
...
+
+  // One-shot Bufferize
```
rainij: This is not a great comment. Could you write a one-line summary of what the pass does (for us).

mlir/test/tools/qcc-opt/jasp-to-qc-conversion-test.mlir
rainij:
```suggestion
// RUN: qcc-opt %s -jasp-to-qc | FileCheck %s
```

mlir/test/tools/qcc/jasp-generate-ghz.mlir
rainij, line 1: Here it would pay off in the future to have a high-level entry point dialect. The input IR is not great to read. Please close thread once read.

mlir/tools/qcc-opt/qcc-opt.cpp
```diff
@@ -1,21 +1,57 @@
...
+  mlir::registerLinalgDetensorizePass();
+  mlir::registerCanonicalizer();
+  mlir::registerCSEPass();
+  mlir::bufferization::registerBufferLoopHoistingPass();
+  mlir::registerMem2RegPass();
+  mlir::registerSCCP();
+  mlir::bufferization::registerPromoteBuffersToStackPass();
+  mlir::registerInlinerPass();
+
+  // Extension registration
```
rainij, line 47: I think the comment does not exactly fit to the code below. The code below contains "external models". I had to look it up, but if I understand it correctly it augments existing dialects (like arith) to know about things they should not depend on (like bufferization interfaces). Right?

IMO it would be sufficient if you just change the line to `// External models`, but if you find a better one ... .

mlir/tools/qcc-opt/qcc-opt.cpp
```diff
@@ -1,21 +1,57 @@
...
+  mlir::arith::registerBufferizableOpInterfaceExternalModels(registry);
+  mlir::linalg::registerBufferizableOpInterfaceExternalModels(registry);
+  mlir::scf::registerBufferizableOpInterfaceExternalModels(registry);
+  mlir::tensor::registerBufferizableOpInterfaceExternalModels(registry);
+  mlir::bufferization::func_ext::registerBufferizableOpInterfaceExternalModels(registry);
+  mlir::memref::registerAllocationOpInterfaceExternalModels(registry);
+  mlir::func::registerInlinerExtension(registry);
```
rainij, line 54:
```suggestion

  // Dialect extensions
  mlir::func::registerInlinerExtension(registry);
```

It does not belong to the group of external models.

mlir/tools/qcc/qcc.cpp
```diff
@@ -25,7 +43,18 @@ int main(int argc, char** argv) {
   cl::ParseCommandLineOptions(argc, argv, "qcc - quantum compiler collection\n");
 
   mlir::DialectRegistry registry;
-  registry.insert<mlir::func::FuncDialect, mlir::arith::ArithDialect, mlir::qc::QCDialect>();
+  registry.insert<mlir::func::FuncDialect, mlir::arith::ArithDialect, mlir::tensor::TensorDialect,
+                  mlir::bufferization::BufferizationDialect, mlir::linalg::LinalgDialect, mlir::scf::SCFDialect,
+                  jasp::JaspDialect, mlir::qc::QCDialect>();
+
+  // Extension registration
```
rainij, line 50: Same remarks apply as in `qcc-opt`.

run_pipeline.sh
rainij, line 1: This was probably added by accident, right?

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -32,13 +39,21 @@ using namespace mlir::qc;
 #include "qcc/Conversion/JaspToQC/JaspToQC.h.inc"
 
 namespace {
-/// Type converter for jasp-to-QC conversion
+/// Type converter for jasp-to-QC conversion, to be used
+/// for op-conversions.
 ///
 /// Handles type conversion between the jasp and QC dialects.
```
rainij, line 45: Not sure how helpful these comments are. I would make the docstring a bit more concise. The class name already says it is a jasp to qc type converter.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij, line 60: Just a question: are these identity conversions really needed?

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -48,6 +63,49 @@ class JaspToQCTypeConverter final : public TypeConverter {
...
+};
+
+/// Type converter for jasp-to-QC conversion, to be
+/// used in function signature conversions.
+///
+/// Handles type conversion between the jasp and QC dialects.
+///  - The jasp qubit type is mapped to the QC qubit type.
+///  - !jasp.QubitArray is mapped to memref<?x!qc.qubit>.
+///  - !jasp.QuantumState is destroyed.
+class QuantumStateEliminator final : public JaspToQCTypeConverter {
```
rainij, line 98: I think the docstring is wrong. I believe it doesn't even need one, I would just remove it.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -48,6 +63,49 @@ class JaspToQCTypeConverter final : public TypeConverter {
...
+public:
+  explicit QuantumStateEliminator(MLIRContext* ctx) : JaspToQCTypeConverter(ctx) {
```
rainij, line 100: Just a question: Why is it even separate from `JaspToQCTypeConverter`? Is there a reason the latter can not do this job?

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -319,6 +363,37 @@ struct ConvertJaspDeleteQubitsOp final : OpConversionPattern<jasp::DeleteQubitsO
...
 };
 
+/// Convert Rank-Zero Tensors to their wrapped types in `linalg.generic` operations.
+/// Only operands are affected.
+struct ConvertRankZeroTensorsInLinalg final : OpConversionPattern<linalg::GenericOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp op, OpAdaptor adaptor,
+                                ConversionPatternRewriter& rewriter) const override {
+    // 1. Collect new input types from the adaptor
```
rainij: These comments look like from gemini :D . At least I always get them when I create code with gemini. Correct me if I am wrong. Especially in this case I would leave out the comment. And I would also leave out the numbers in all the comments.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -319,6 +363,37 @@ struct ConvertJaspDeleteQubitsOp final : OpConversionPattern<jasp::DeleteQubitsO
...
+    // 2. Update indexing maps if any tensor became a scalar
+    SmallVector<AffineMap> newMaps = op.getIndexingMapsArray();
+    for (auto it : llvm::enumerate(op.getInputs())) {
+      if (isa<RankedTensorType>(it.value().getType()) &&
+          llvm::cast<RankedTensorType>(it.value().getType()).getRank() == 0) {
+        newMaps[it.index()] = rewriter.getMultiDimIdentityMap(0); // () -> ()
+      }
+    }
+
+    // 3. Create the new generic op
```
rainij: No helpful.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -319,6 +363,37 @@ struct ConvertJaspDeleteQubitsOp final : OpConversionPattern<jasp::DeleteQubitsO
...
+    auto newOp = linalg::GenericOp::create(rewriter, op.getLoc(), op.getResultTypes(), newInputs, adaptor.getOutputs(),
+                                           newMaps, op.getIteratorTypesArray());
+
+    // 4. Move the region over
```
rainij: Cool feature, but I believe one can still remove the comment. The line of code stands for itself.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -319,6 +363,37 @@ struct ConvertJaspDeleteQubitsOp final : OpConversionPattern<jasp::DeleteQubitsO
...
+    rewriter.inlineRegionBefore(op.getRegion(), newOp.getRegion(), newOp.getRegion().begin());
+
+    rewriter.replaceOp(op, newOp.getResults());
+    return success();
+  }
+};
+
 /// Pass implementation for jasp-to-QC conversion
 ///
 /// This pass converts the Jasp dialect to the QC dialect. It handles the
```
rainij: I think we should consider not adding a docstring to the pass implementations. The tablegen file should be the source of truth for this class IMO.

rainij (changes requested):
Very cool, contains alot of features in MLIR we did not look into so far. I have a bunch of minor change requests only.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: Yes. MLIR wants to be able to look up any type it encounters. So one has to either make this very explicit, and list all the types appearing in the IR, or use this fallback. There's an error if I remove this line.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: I corrected the docstring, it makes sense to have some explanation IMO as the code is a bit strange.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: I could just not get it to work with a single converter and gave up after some time. I don't understand why. There's a comment that this is a hack in the `JaspToQCTypeConverter` docstring.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: Here and in the pipeline construction I was very lazy and should have done some cleanup. Removed the comments here and in some other places, as I am also not a fan of this redundant explanatory style.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: True, removed

run_pipeline.sh
julfarn: Ah yes, removed now

mlir/tools/qcc-opt/qcc-opt.cpp
julfarn: I'm not sure. In my view, external models are extensions. They introduce some interfaces to dialects. This is the error message if you forget the registration: 

```
LLVM ERROR: checking for an interface (`mlir::bufferization::BufferizableOpInterface`) that was promised by dialect 'arith' but never implemented. This is generally an indication that the dialect extension implementing the interface was never registered.
```

It's the same for extensions which are officially called extensions, like the inliner extension of the func dialect.

mlir/tools/qcc-opt/qcc-opt.cpp
julfarn: See other comment, I think these make a common group.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: True, do we have a guideline here? I am following the style of a one-line summary, then a description. But the summary is often redundant.

Here, there's the additional caveat that this class is only to be used for op-conversions, not type conversions in signatures.

mlir/lib/Compiler/Pipeline.cpp
julfarn: I added a lot of comments explaining the pipeline. This is a case where explaining every step actually makes sense in my opinion, and I should have put more care into this from the beginning.

mlir/lib/Compiler/Pipeline.cpp
julfarn: rewritten

mlir/tools/qcc/qcc.cpp
julfarn: I believe the term is appropriate, see other comment

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -1,22 +1,73 @@
...
+
+  // Initial cleanup
+  pm.addNestedPass<mlir::func::FuncOp>(mlir::createCanonicalizerPass());
+  pm.addNestedPass<mlir::func::FuncOp>(mlir::createCSEPass());
+
+  // Convert `tensor.empty` to `bufferization.alloc_tensor`, which is expected by
+  // bufferization.
+  pm.addPass(mlir::bufferization::createEmptyTensorToAllocTensorPass());
+
+  // Detensorization attempts to converts those rank-0-tensors to plain values
```
burgholzer:
```suggestion
  // Detensorization attempts to convert those rank-0-tensors to plain values
```

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -1,26 +1,33 @@
...
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypeInterfaces.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/OperationSupport.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/IR/Types.h"
+#include "mlir/IR/ValueRange.h"
+#include "mlir/Support/LLVM.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Transforms/DialectConversion.h"
```
burgholzer, line 23: Why was the import style changed here? My personal impression was that external imports (which all non-project MLIR/LLVM imports are) should be included using `<...>` syntax in general.
I am definitely not hard set on that, but I think it should be a conscious decision.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -32,13 +39,21 @@ using namespace mlir::qc;
...
 /// Handles type conversion between the jasp and QC dialects.
 ///  - The jasp qubit type is mapped to the QC qubit type.
 ///  - !jasp.QubitArray is mapped to memref<?x!qc.qubit>.
-///  - TODO: !jasp.QuantumState is destroyed.
-class JaspToQCTypeConverter final : public TypeConverter {
+///  - !jasp.QuantumState is left intact.
+///    TODO: This is a hack. If we delete the QuantumState
+///    type here, conversion fails before any pattern is applied.
+///    But if we leave it out, function signatures will not be
+///    updated. Therefore, we need two separate type converters.
```
burgholzer, line 52: Should a tracking issue be created for this? 
I have no solution for the underlying problem in mind at the moment, but it feels like something that we should resolve at some point.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -319,12 +354,33 @@ struct ConvertJaspDeleteQubitsOp final : OpConversionPattern<jasp::DeleteQubitsO
...
+    SmallVector<AffineMap> newMaps = op.getIndexingMapsArray();
+    for (auto it : llvm::enumerate(op.getInputs())) {
+      if (isa<RankedTensorType>(it.value().getType()) &&
+          llvm::cast<RankedTensorType>(it.value().getType()).getRank() == 0) {
+        newMaps[it.index()] = rewriter.getMultiDimIdentityMap(0);
+      }
+    }
+
+    auto newOp = linalg::GenericOp::create(rewriter, op.getLoc(), op.getResultTypes(), newInputs, adaptor.getOutputs(),
+                                           newMaps, op.getIteratorTypesArray());
```
burgholzer:
```suggestion
    SmallVector<AffineMap> newMaps = op.getIndexingMapsArray();
    for (auto it : llvm::enumerate(op.getInputs())) {
      if (isa<RankedTensorType>(it.value().getType()) &&
          llvm::cast<RankedTensorType>(it.value().getType()).getRank() == 0) {
        newMaps[it.index()] = rewriter.getMultiDimIdentityMap(0);
      }
    }

    auto newOp = linalg::GenericOp::create(rewriter, op.getLoc(), op.getResultTypes(), adaptor.getInputs(), adaptor.getOutputs(),
                                           newMaps, op.getIteratorTypesArray());
```
I am pretty sure there is no need to materialize a vector here and one can simply use the range returned by `adaptor.getInputs()`.

burgholzer (approved):
This looks pretty good to me!
All my comments are on the nitpick side of things.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij: I think we can have a dedicated PR on this. My PR also had an issue with the tooling (I could not commit and CI contradicted itself) with regard to includes.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij: No not really for fullstaqd. I just have Valentin in my ear on that. But it is a small detail, so please decide yourself what is best (and close thread afterwards).

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij: OK, I see. But it isn't so bad anyway IMO.

mlir/lib/Compiler/Pipeline.cpp
rainij: Awesome!

mlir/test/tools/qcc-opt/jasp-to-qc-conversion-test.mlir
rainij: @julfarn would you agree to this format? just one "-". So far (also in our internal project) we used it with just one "-". Might become a small guideline (unwritten).

mlir/tools/qcc-opt/qcc-opt.cpp
rainij: I have the feeling that the external models are more lightweight. It makes sense that they have the same error message for both. I believe the error message cannot know in which way you implemented it. I guess they just chose the most general solution.

rainij (approved):
@julfarn I approved to not block this PR for too long (I am away for some hours today). But could you go over my remaining open threads and decide for yourself what makes sense. After that feel free to merge yourself.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: I feel like leaving it in for now, but in principle I agree with you (and Valentin)

mlir/tools/qcc-opt/qcc-opt.cpp
julfarn: The pipeline will undoubtedly change in the future, let's revisit this then.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: > My PR also had an issue with the tooling (I could not commit and CI contradicted itself) with regard to includes.

Yes exactly, I was fighting the CI on the import ordering. My preference would also be to fix it later

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: Agreed. Opened #36 on this.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: @burgholzer, please merge if you think this is acceptable.

burgholzer: Yeah. Let's address this in a follow up 👍🏼

PR #37, opened 2026-04-15
Update clang tidy and format

rainij:
The update mainly affects the include order and follows llvm coding conventions. There are also a few minor updates to clang-tidy config which make sense for an llvm project. The current clang format config would order things like this:

```cpp
// inside qcc.cpp
#include "qcc.h" // This has implicitly priority 0 as this file is qcc.cpp

#include "qcc/Compiler/Pipeline.h"
#include "qcc/Dialect/Jasp/IR/Jasp.h"

#include "bar.h"
#include "foo.h"

#include "mlir/Conversion/VectorToSCF/VectorToSCF.h"
#include "mlir/Dialect/Arith/IR/Arith.h"
#include "mlir/Dialect/Arith/Transforms/BufferizableOpInterfaceImpl.h"

#include "llvm/Support/CommandLine.h"
#include "llvm/Support/SourceMgr.h"
#include "llvm/Support/ToolOutputFile.h"

#include <std_header.h>
#include <system_header.h>
```

The order is as suggested by the llvm style guide (although it does not mention 3rd party libs because it hasn't any dependencies).

I did a bit of research on `<` (angle includes) vs `"` (quote includes). It seems that especially for llvm projects but also in "modern C++ development" the rules are as follows

- Use angles for system and os headers.
- Rarely angles are used for "particularly" stable third party libraries.
- Use quotes for all the rest: own headers, llvm headers, 3rd party lib headers.

I have no good reference on this but if the need arises we can definitely question this and try to come up with good references. For me what matters is that we stick to the established way to do it - whatever that is. It would probably good to know this anyway, but I want anybodies feedback first.

Things the current config does not resolve in all edge cases:
- "private" headers like `foo.h` would be lumped together with 3rd party libs. But this could in principle be resolved by explicitly introducing rules for each 3rd party lib we depend on - if this is desired and does not get cumbersome.
- If somebody accidentally uses the wrong style (angle vs quotes) then `<llvm/blah.h>` would count as system header.


rainij:
@burgholzer @julfarn please let us start a discussion on this topic. Today I am in the mood of finding the perfect solution for such things :D . Note that I only changed the two config files, the rest was the CI.

burgholzer:
LLVM itself only really specifies the order of the particular includes in its coding guidelines https://llvm.org/docs/CodingStandards.html#include-style

The Google C++ Guide has the following interesting point:
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

> Headers should only be included using an angle-bracketed path if the library requires you to do so. In particular, the following headers require angle brackets:
> - C and C++ standard library headers (e.g., <stdlib.h> and <string>).
> - POSIX, Linux, and Windows system headers (e.g., <unistd.h> and <windows.h>).
> - In rare cases, third_party libraries (e.g., <Python.h>).

I am fine to take this as justification for going with `"..."` over `<...>` for the LLVM/MLIR includes.

@denialhaag for awareness.

denialhaag:
Regarding four of the linter errors, `misc-const-correctness` is disabled in the MLIR project: https://github.com/llvm/llvm-project/blob/70cf763a42d55dcf89687c9e84478eb007561654/mlir/.clang-tidy#L3

The current `llvm::cl::opt` usage aligns with many LLVM/MLIR examples.

rainij:
> The Google C++ Guide has the following interesting point:
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

@burgholzer nice something like this I was looking for! I would like to make sure that we do not lose track of this justification. Should we put the link and the quote into the clang format config file so that we are aware of this. Or does it fit better into one of the repo files like `CONTRIBUTING.md`.

rainij:
> Regarding four of the linter errors, misc-const-correctness is disabled in the MLIR project: https://github.com/llvm/llvm-project/blob/70cf763a42d55dcf89687c9e84478eb007561654/mlir/.clang-tidy#L3

@denialhaag in our private project we also disabled this check. It never felt unnatural although I was previously following this rule in any other C++ related project. I think for a MLIR projects it is totally justified. A compiler is about mutating something all the time (the IR) and I guess it would just add lots of boilerplate to add support for const correctness. Alot of stuff also behaves like a smart pointer.

IMO we can totally follow the MLIR lead for this project. I guess you are in favour @denialhaag . What about @burgholzer ?

burgholzer:
> > The Google C++ Guide has the following interesting point:
> > https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
> 
> @burgholzer nice something like this I was looking for! I would like to make sure that we do not lose track of this justification. Should we put the link and the quote into the clang format config file so that we are aware of this. Or does it fit better into one of the repo files like `CONTRIBUTING.md`.

I don't have particularly strong feelings on where to put this. Maybe both? In the clang-format config file because this is what the rules are based on; and in the contributing docs so that potential collaborators also directly know about it.

burgholzer:
> > Regarding four of the linter errors, misc-const-correctness is disabled in the MLIR project: https://github.com/llvm/llvm-project/blob/70cf763a42d55dcf89687c9e84478eb007561654/mlir/.clang-tidy#L3
> 
> @denialhaag in our private project we also disabled this check. It never felt unnatural although I was previously following this rule in any other C++ related project. I think for a MLIR projects it is totally justified. A compiler is about mutating something all the time (the IR) and I guess it would just add lots of boilerplate to add support for const correctness. Alot of stuff also behaves like a smart pointer.
> 
> IMO we can totally follow the MLIR lead for this project. I guess you are in favour @denialhaag . What about @burgholzer ?

I am also fine with following the MLIR style here (despite being a const-advocate in all other C++ projects we do).
Some justification that we may also link: https://mlir.llvm.org/docs/Rationale/UsageOfConst/

rainij:
Then I think now we probably have a final version. Today I changed the following:
- Document include style in .clang-format and `CONTRIBUTING.md`.
- Add C++ guidelines to `CONTRIBUTING.md` (I tried to make it concise but still informative).
- Update clang tidy on dropping the const correctness check.

.clang-format
```diff
@@ -1,12 +1,30 @@
 BasedOnStyle: LLVM
 ColumnLimit: 120
+
+# Handling of include blocks must be consistent with clang-tidy(llvm-include-order).
+# See also: https://llvm.org/docs/CodingStandards.html#include-style.
+#
+# NOTE: Headers should only be included using an angle-bracketed path of the library requires you
```
burgholzer:
```suggestion
# NOTE: Headers should only be included using an angle-bracketed path if the library requires you
```

CONTRIBUTING.md
```diff
@@ -40,6 +40,35 @@ Please adhere to the following guidelines to help the project grow sustainably.
 - Follow the project's coding standards and conventions.
 - Be open to feedback and willing to make necessary changes based on code reviews.
 
+### C++ guidelines
+
+As an LLVM/MLIR base project we mostly follow their guidelines.
+We also follow googles guidelines but LLVM guidelines take precedence.
```
burgholzer:
```suggestion
We also follow Google's guidelines but LLVM guidelines take precedence.
```

CONTRIBUTING.md
```diff
@@ -40,6 +40,35 @@ Please adhere to the following guidelines to help the project grow sustainably.
...
+
+As a particular example, we follow [LLVMs recommendation](https://llvm.org/docs/CodingStandards.html#include-style)
+on include order from most specific (top) to least specific (bottom).
+
+When it comes to includes with angle brackets vs quoted includes we follow google's recommendation:
```
burgholzer:
```suggestion
When it comes to includes with angle brackets vs quoted includes we follow Google's recommendation:
```

CONTRIBUTING.md
```diff
@@ -40,6 +40,35 @@ Please adhere to the following guidelines to help the project grow sustainably.
...
+
+- C and C++ standard library headers (e.g., <stdlib.h> and ).
+- POSIX, Linux, and Windows system headers (e.g., <unistd.h> and <windows.h>).
+- In rare cases, third_party libraries (e.g., <Python.h>).
+
+See also our `.clang-format` file.
+
+Also note that MLIR based projects typically do not _strictly_ follow const correctness.
+The details can be found in [MLIR on the usage of 'const'](https://mlir.llvm.org/docs/Rationale/UsageOfConst/).
+The slightly simpliefied gist is as follows:
```
burgholzer:
```suggestion
The slightly simplified gist is as follows:
```

burgholzer (approved):
LGTM besides some typos. You should be able to quickly batch these up.

julfarn:
I have nothing to add, looks reasonable.

julfarn (approved):

PR #38, opened 2026-04-15
Add simple .clangd

rainij:
Just a proposal. The alternative would be to let everybody set such stuff locally. Opinions welcome. I am open to close the PR without merging. Then we would at least have discussed it.

Rationale for not flagging unused includes is that in MLIR we get a lot of false positives due to these generated headers `*.h.inc` actually requiring some other headers before them.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,2 @@
+Diagnostics:
+  UnusedIncludes: None


burgholzer (approved):
I am not entirely opposed to this. The include tracker has many false positives especially with LLVM/MLIR related code.
At the same time, we also disabled the include cleaner clang-tidy check, which would generally reveal missing includes (but also has its own false positives that are annoying).

As long as we have broad enough CI coverage (and we currently do; covering all major operating systems and all major compilers), I am fine with including this. As soon as we notice any friction in subsequent PRs (for example by relying on transitive includes), we should remove this again.

The only thing that has me worried a little bit is that there is now no mechanism except for human reviews that would point out superfluous includes. Given how these can have a non-negligible impact on compile time, I am questioning whether it is really too bad to add a couple of `// IWYU pragma: keep` for the false positives.
(https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md for those not aware of the existence of this handy tool).

burgholzer:
Probably also worth to keep https://llvm.org/docs/CodingStandards.html#include-as-little-as-possible in mind

Edit: Also https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

rainij:
> The only thing that has me worried a little bit is that there is now no mechanism except for human reviews that would point out superfluous includes. Given how these can have a non-negligible impact on compile time, I am questioning whether it is really too bad to add a couple of // IWYU pragma: keep for the false positives.
(https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md for those not aware of the existence of this handy tool).

@burgholzer Actually I did not know about this. And I also share your worries about the "true positives". It is really hard to spot unused imports by hand. So I tend towards closing the PR without merging. We can share this IWYU pragma thing on discord so that everybody knows (my team certainly does not).

I will just wait for @julfarn to comment on that before actually closing the PR.

julfarn:
Noted. I also rely on tooling to detect unused imports, and marking them is actually beneficial I believe. Thanks for pointing this out.

rainij:
Closing, I shared the IWYU trick with my team.

PR #43, opened 2026-05-07
Checking constant allocation sizes. Removing memref dialect usage.

RiccardoRomanello:
Added two passes into JaspToQC directory to:
- check compilation time sizes allocation
- remove memref usage


mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,36 @@
...
+      errors for non-constant or invalid sizes.
+   }];
+  let dependentDialects = [
+    "mlir::qc::QCDialect",
+    "mlir::tensor::TensorDialect",
+    "mlir::memref::MemRefDialect"
+  ];
+}
+
+def StaticizeQubitRefs : Pass<"staticize-qubit-refs"> {
```
rainij: The name of the pass is a bit awkward. Maybe `convert-to-static-qubits`? @burgholzer any oppinion?

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,36 @@
...
+  let summary = "Replace d ynamic qubit array loads with static qubit references.";
```
```suggestion
  let summary = "Replace dynamic qubit array loads with static qubits.";
```

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,36 @@
...
+                    let description = [{
+      Transforms memref.load on qubit arrays into direct qc.static references.
```
```suggestion
      Transforms `memref.load` on qubit arrays into direct `qc.static` references.
```

rainij: On the other hand I actually would avoid to repeat the summary again. I think if we would render it in html this would look repetitive.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,36 @@
...
+      Requires compile-time constant indices (typically from loop unrolling).
+   }];
+  let dependentDialects = [
+    "mlir::qc::QCDialect",
+    "mlir::tensor::TensorDialect",
+    "mlir::memref::MemRefDialect"
```
rainij: Can you check if which ones you need. Typically you only list those dialects for which you create _new_ ops. But you do not _create_ e.g. memref stuff.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
```diff
@@ -0,0 +1,84 @@
...
+namespace {
+
+struct CheckStaticQubitAllocation final : public impl::CheckStaticQubitAllocationBase<CheckStaticQubitAllocation> {
+  using CheckStaticQubitAllocationBase<CheckStaticQubitAllocation>::CheckStaticQubitAllocationBase;
+
+protected:
+  /**
+    Ensures qubit memory allocations use positive, compile-time constant sizes. This pass prevents dynamic qubit
+    scaling, which is unsupported by the target quantum backend.
+  */
```
rainij: Please remove the documentation from here. Our single source of truth (for docs of passes) should be the corresponding tablegen file. Please add all relevant details there.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,36 @@
 include "mlir/Pass/PassBase.td"
 
+def CheckStaticQubitAllocation : Pass<"check-static-qubit-allocation"> {
+  let summary = "Validate that qubit array allocations use compile-time constant sizes.";
+  let description = [{
+      Verifies all qubit memref allocations have static sizes and reports diagnostic
+      errors for non-constant or invalid sizes.
```
rainij, line 16: Please add an IR example which is OK in terms of this pass and one which would be flagged as error.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
```diff
@@ -0,0 +1,84 @@
...
+      }
+
+      if (memrefType.isDynamicDim(0)) {
+        allocOp->emitError("qubit array allocation size must be a compile-time constant; "
+                           "dynamic qubit array sizes are not supported");
+        failed = true;
+        return;
+      }
+
+      int64_t size = memrefType.getDimSize(0);
```
rainij: Personal taste: I would use `auto` here because the function `getDimSize` already decides about the datatype.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
```diff
@@ -0,0 +1,84 @@
...
+  using CheckStaticQubitAllocationBase<CheckStaticQubitAllocation>::CheckStaticQubitAllocationBase;
+
+protected:
+  /**
+    Ensures qubit memory allocations use positive, compile-time constant sizes. This pass prevents dynamic qubit
+    scaling, which is unsupported by the target quantum backend.
+  */
+  void runOnOperation() override {
+    Operation* op = getOperation();
+    bool failed = false;
```
rainij: Please do not use these flag variables like `failed`. In the walk below use instead `return WalkResult::interrupt()` (failure) and `return WalkResult::advance()` (normal case) and retrieve the corresponding `WalkResult result = op->walk(...);`.

Then below you can just do

```cpp
if (result.wasInterrupted()) {
  signalPassFailure();
}
```

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
```diff
@@ -0,0 +1,84 @@
...
+    op->walk([&](memref::AllocOp allocOp) {
+      auto memrefType = dyn_cast<MemRefType>(allocOp.getType());
+
+      if (!memrefType || !isa<qc::QubitType>(memrefType.getElementType())) {
+        return;
+      }
+
+      if (memrefType.isDynamicDim(0)) {
+        allocOp->emitError("qubit array allocation size must be a compile-time constant; "
+                           "dynamic qubit array sizes are not supported");
```
rainij:
```suggestion
        allocOp.emitError("qubit array allocation size must be a compile-time constant; "
                           "dynamic qubit array sizes are not supported");
```

rainij: Fits with general MLIR coding style to use the method from the typed handle (`AllocOp`) instead of the underlying `Operation*`.

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/OperationSupport.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/IR/Types.h"
+#include "mlir/IR/ValueRange.h"
+#include "mlir/Support/LLVM.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+#include "llvm/ADT/DenseMap.h" // DenseMap
```
rainij:
```suggestion
#include "llvm/ADT/DenseMap.h"
```

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "llvm/Support/Casting.h"
+
+#include <cassert>
+#include <utility> // std::pair
```
rainij:
```suggestion
#include <utility>
```

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
```diff
@@ -0,0 +1,149 @@
...
+    Operation* op = getOperation();
+    OpBuilder builder(op->getContext());
+
+    // Tracks the mapping from a (MemRef, Index) pair to a specific physical qubit Value.
+    // This essentially "flattens" the memory model into a lookup table.
+    DenseMap<std::pair<Value, int64_t>, Value> qubitMap;
+
+    // A global counter to ensure every physical qubit in the circuit gets a unique hardware ID.
+    int64_t nextGlobalQubitIdx = 0;
+    bool failed = false;
```
rainij: Please avoid `failed` you can use similar tricks as the ones I mentioned in the other pass.

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
```diff
@@ -0,0 +1,149 @@
...
+      loadOp.getResult().replaceAllUsesWith(it->second);
+      loadOp.erase(); // The load is now redundant.
+    });
+
+    if (failed) {
+      return signalPassFailure();
+    }
+
+    // Note: memref.alloc and dealloc ops are typically cleaned up by
+    // subsequent standard MLIR buffer-deallocation or Canonicalization passes.
```
rainij: Remove this comment from here and instead document this as an assumption by this pass to the tablegen.

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
```diff
@@ -0,0 +1,149 @@
...
+    // Tracks the mapping from a (MemRef, Index) pair to a specific physical qubit Value.
+    // This essentially "flattens" the memory model into a lookup table.
+    DenseMap<std::pair<Value, int64_t>, Value> qubitMap;
+
+    // A global counter to ensure every physical qubit in the circuit gets a unique hardware ID.
+    int64_t nextGlobalQubitIdx = 0;
+    bool failed = false;
+
+    // Helper: Identifies if a type is a MemRef containing Qubits.
+    auto isQubitMemref = [](Type type) {
```
rainij: It looks like this could be a plain old-style helper function.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
rainij: It might be an option to give an IR example. You should explain in a bit more detail how this works.

mlir/test/tools/qcc/ghz_from_qrisp_v2.mlir
rainij, line 1: Why are there two versions of the file?

mlir/test/tools/qcc/ghz_from_qrisp.mlir
rainij, line 1: Please also add tests for `qcc-opt` which test the two passes individually.

mlir/test/tools/qcc/ghz_from_qrisp.mlir
```diff
@@ -0,0 +1,120 @@
...
+    %11 = "jasp.quantum_gate"(%7, %10, %8) {gate_type = "cx"} : (!jasp.Qubit, !jasp.Qubit, !jasp.QuantumState) -> !jasp.QuantumState
+    %12, %13 = "jasp.measure"(%4, %11) : (!jasp.Qubit, !jasp.QuantumState) -> (tensor<i1>, !jasp.QuantumState)
+    %14, %15 = "jasp.measure"(%7, %13) : (!jasp.Qubit, !jasp.QuantumState) -> (tensor<i1>, !jasp.QuantumState)
+    %16, %17 = "jasp.measure"(%10, %15) : (!jasp.Qubit, !jasp.QuantumState) -> (tensor<i1>, !jasp.QuantumState)
+
+    %m0 = tensor.extract %12[] : tensor<i1>
+    %m1 = tensor.extract %14[] : tensor<i1>
+    %m2 = tensor.extract %16[] : tensor<i1>
+
+    aux.record_bool %m0
```
rainij, line 57: How was this file created? I cannot imagine that qrisp already creates `aux.record_bool`.

rainij (changes requested):

rainij:
@RiccardoRomanello if the PR is related or even resolves an issue, please mention it in the description of the PR.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
burgholzer: I agree that the current name is a bit odd. I like your suggestion @rainij.i was briefly thinking about whether it is clear "from what" the conversion to static qubits happens and if that should be part of the pass name. But it's probably fine as-is.

rainij: Yes I have the same feeling. The name I proposed seems better than the current one but there is room for improvement. I will have a second thought in my next review.

RiccardoRomanello: What about memref-to-static-qc?
There is a clear idea of starting and ending point

rainij: _Maybe_ could make sense to have the `convert-` prefix. I am not sure though ... . I would need to have a second view on it.

burgholzer: Maybe `convert-memref-to-static-qubits`?

RiccardoRomanello: I pushed after applying the first round of reviews without changing the name. Once we get to a fix point, I will change all the required components.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: Verified that actually for one of the two steps, no dialect was required. Thanks for the insights about the _create_ function.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
RiccardoRomanello: Done

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: Added both a working and a non-working example. For the former one, I also added 'what to expect' as an output.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
RiccardoRomanello: Done

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
RiccardoRomanello: I replaced the `failed` variable as u suggested. The walks now correctly rely on `WalkResult` to communicate their outcome.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
RiccardoRomanello: Thanks for the insight, will keep that in mind.

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
RiccardoRomanello: As mentioned above, removed any dependencies to the _failed_ variable

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
RiccardoRomanello: Removed all the _documentation_ like comments from the files and moved them to the `td`

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
RiccardoRomanello: Declared a _proper_ function

mlir/test/tools/qcc/ghz_from_qrisp_v2.mlir
RiccardoRomanello: My mistake, sorry

mlir/test/tools/qcc/ghz_from_qrisp.mlir
RiccardoRomanello: Added two files to test the `qcc-opt` tool as well.

mlir/test/tools/qcc/ghz_from_qrisp.mlir
RiccardoRomanello: I actually edited by hand the result of qrisp to use the aux dictionary as well. It seemed to me that it would be more aligned with the example you gave of GHZ.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,49 @@
...
+       %2 = qc.static 2 : !qc.qubit`
+      in which the allocation of three qubits has been turned into 3 separate static allocations.
+
+      On the other hand, in the following snippet of code:
+      `func.func public @main(%arg14: !jasp.QuantumState, %s: tensor<i64>) attributes { qcc.entry_point } {
+        %1, %2 = "jasp.create_qubits"(%s, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)`
+      the compilation process will prompt an error since the number of element that `jasp.create_qubits` needs to allocate
+      is unknown at compile time.
+
+      The compiler will prompt an error message even if the size of the register we want to allocate is negative.
```
EnriqueNB:
```suggestion
      The compiler will prompt an error message even if the size of the register we want to allocate is negative, indicating dynamic memory allocation.
```
Is using a negative size for register a standard way for indicating that dynamic memory allocation is expected. If so, I would add an extra comment in the description as it was obscure to me.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,49 @@
...
+   }];
+}
+
+def StaticizeQubitRefs : Pass<"staticize-qubit-refs"> {
+  let summary = "Replace dynamic qubit array loads with static qubits.";
+  let description = [{ First, substitutes `memref.alloc` with a list of `qc.static` allocation.Subsequently,
+      transforms `memref.load` on qubit arrays into direct `qc.static` references.Requires compile -
+          time constant indices(should be called after a `check - static - qubit - allocation` step)
+              .It assumes dangling `memref.load` calls to be cleaned up via further `mlir::Canonicalization`.
+}];
```
EnriqueNB:
```suggestion
  let description = [{ First, substitutes `memref.alloc` with a list of `qc.static` allocation. Subsequently,
       transforms `memref.load` on qubit arrays into direct `qc.static` references. Requires compile-time
       constant indices(should be called after a `check-static-qubit-allocation` step). It assumes dangling
       `memref.load` calls to be cleaned up via further `mlir::Canonicalization`.
}];
```
I changed the formatting.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,49 @@
...
+let dependentDialects = ["mlir::qc::QCDialect"];
+}
+
 def JaspToQC : Pass<"jasp-to-qc"> {
   let summary = "Convert jasp dialect to QC dialect.";
 
   let description = [{
-      This pass converts all operations from qrisp's jasp dialect
+      This pass converts all operations from qrisps jasp dialect
       (see https://qrisp.eu/reference/Jasp/MLIR%20Interface.html)
```
EnriqueNB, line 50: Here I would explicitly put the operations from jasp and their equivalent in the QC dialect. There are only three operations and that website may change. Let document what we have in the code and not point to a website that can change.

mlir/lib/Conversion/JaspToQC/JaspCheckStaticQubitAllocation.cpp
```diff
@@ -0,0 +1,80 @@
...
+        return WalkResult::interrupt();
+      }
+
+      return WalkResult::advance();
+    });
+
+    if (result.wasInterrupted()) {
+      signalPassFailure();
+    }
+  }
```
EnriqueNB, line 58:
```suggestion
    if (op->walk([&](memref::AllocOp allocOp) {
  auto memrefType = cast<MemRefType>(allocOp.getType());
  if (!isa<qc::QubitType>(memrefType.getElementType()))
    return WalkResult::advance();

  int64_t size = memrefType.getDimSize(0);
  if (size > 0)
    return WalkResult::advance();

  auto diag = allocOp.emitError("qubit array size must be ");
  memrefType.isDynamicDim(0) ? diag << "a compile-time constant" 
                             : diag << "positive, got " << size;

  return WalkResult::interrupt();
}).wasInterrupted()) signalPassFailure();
```
Here is a condensed version (and I think more clear).

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
```diff
@@ -0,0 +1,147 @@
...
+  using StaticizeQubitRefsBase<StaticizeQubitRefs>::StaticizeQubitRefsBase;
+
+private:
+  // Identifies if a type is a MemRef containing Qubits.
+  static auto isQubitMemref(Type type) {
+    if (auto mType = dyn_cast<MemRefType>(type)) {
+      return isa<qc::QubitType>(mType.getElementType());
+    }
+
+    return false;
```
EnriqueNB:
```suggestion
  static bool isQubitMemref(Type type) {
    auto mType = dyn_cast<MemRefType>(type);
    return mType && isa<qc::QubitType>(mType.getElementType());
  }
```
This part can be simplified. Also we know that return type is a bool so can explicitly annotate the function.

EnriqueNB (changes requested):
I added a few comments about the formatting and simplification of the code. The logic look good to me. Please, consider them.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
rainij: Indeed the sentence sounds a bit weird. Actually I would be super surprised if negative values mean anything in this context.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
rainij: Then let us use `convert-memref-to-static-qubits`.

mlir/lib/Conversion/JaspToQC/JaspCheckStaticQubitAllocation.cpp
rainij: I am not sure if this is really clearer. Goes a bit in the direction of code golf IMO. I think the two steps, first `WalkResult result = ...;` then `if (...) ...` is actually more readable.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
```diff
@@ -0,0 +1,80 @@
...
+      if (size <= 0) {
+        allocOp.emitError("qubit array size must be positive, got ") << size;
+        return WalkResult::interrupt();
+      }
+
+      return WalkResult::advance();
+    });
+
+    if (result.wasInterrupted()) {
+      signalPassFailure();
```
rainij:
```suggestion
      return signalPassFailure();
```

This is a standard pattern I see often in LLVM code. Here nothing follows, but I think one should not go on anyway after a pass failure (most of the time). This explicit return signals the intend.

Also: Some linter recommend two lines for this (`signalPassFailure(); return;`). The typical argument is that one should not return `void`. But IMO it is more readable (needs less lines of code and is still not convoluted). Moreover this would follow standard LLVM patterns.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
```diff
@@ -0,0 +1,80 @@
+
```
rainij, line 9: Redundant newline.

mlir/lib/Conversion/JaspToQC/JaspCheckStaticQubitAllocation.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <utility>
+
+namespace qcc {
+
+#define GEN_PASS_DEF_CHECKSTATICQUBITALLOCATION
+#include "qcc/Conversion/JaspToQC/JaspToQC.h.inc"
+
+using namespace jasp;
+using namespace mlir;
+using namespace mlir::qc;
```
rainij, line 22:
```suggestion
#include "qcc/Conversion/JaspToQC/JaspToQC.h"

#include "mlir/Dialect/MemRef/IR/MemRef.h"
#include "mlir/Dialect/QC/IR/QCDialect.h"
#include "mlir/Dialect/QC/IR/QCOps.h"
#include "mlir/Support/LLVM.h"

namespace qcc {

#define GEN_PASS_DEF_CHECKSTATICQUBITALLOCATION
#include "qcc/Conversion/JaspToQC/JaspToQC.h.inc"

using namespace mlir;
using namespace mlir::qc;
```

rainij: I tried it locally, most headers seem not be relevant. Please also check personally due to two reasons:

- I did not check if even more headers are redundant (maybe not).
- I asked Claude to come up with this, it might have been too aggressive in the sense that one include only appears redundant because it is still transitively included. In that case it is best practice to leave it in.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
rainij: Looks almost good now, the declaration of `failed` is still present in the source code though. Apparently the linter does not flag unused variables.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <utility>
+
+namespace qcc {
+
+#define GEN_PASS_DEF_STATICIZEQUBITREFS
+#include "qcc/Conversion/JaspToQC/JaspToQC.h.inc"
+
+using namespace jasp;
+using namespace mlir;
+using namespace mlir::qc;
```
rainij, line 30:
```suggestion
#include "qcc/Conversion/JaspToQC/JaspToQC.h"

#include "mlir/Dialect/MemRef/IR/MemRef.h"
#include "mlir/Dialect/QC/IR/QCDialect.h"
#include "mlir/Dialect/QC/IR/QCOps.h"
#include "mlir/IR/Builders.h"
#include "mlir/IR/PatternMatch.h"
#include "mlir/IR/Types.h"
#include "mlir/Support/LLVM.h"

#include "llvm/ADT/DenseMap.h"

#include <utility>

namespace qcc {

#define GEN_PASS_DEF_STATICIZEQUBITREFS
#include "qcc/Conversion/JaspToQC/JaspToQC.h.inc"

using namespace mlir;
using namespace mlir::qc;
```

rainij: And check this similar to the change in the other pass.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+namespace {
+
+struct StaticizeQubitRefs final : public impl::StaticizeQubitRefsBase<StaticizeQubitRefs> {
+  using StaticizeQubitRefsBase<StaticizeQubitRefs>::StaticizeQubitRefsBase;
+
+private:
+  // Identifies if a type is a MemRef containing Qubits.
```
rainij:
```suggestion
  /// Identifies if a type is a MemRef containing Qubits.
```

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
```diff
@@ -0,0 +1,147 @@
...
+      // Retrieve the pre-allocated physical qubit from our map.
+      auto it = qubitMap.find({baseMemref, idx});
+      if (it == qubitMap.end()) {
+        loadOp->emitError("internal error: static qubit not found for this allocation");
+        return WalkResult::interrupt();
+      }
+
+      // Replace all uses of the 'loaded' qubit with the 'static' hardware qubit.
+      loadOp.getResult().replaceAllUsesWith(it->second);
+      loadOp.erase(); // The load is now redundant.
```
rainij:
```suggestion
      loadOp.erase();
```

mlir/test/tools/qcc-opt/jasp-check-static-qubit-allocation-test.mlir
```diff
@@ -0,0 +1,74 @@
...
+    } : !qc.qubit
+    %3 = qc.measure %0 : !qc.qubit -> i1
+    %4 = qc.measure %1 : !qc.qubit -> i1
+    %5 = qc.measure %2 : !qc.qubit -> i1
+    aux.record_bool %3
+    aux.record_bool %4
+    aux.record_bool %5
+    return
+  }
+
```
rainij, line 8: I would add a comment that the output should be (essentially) unchanged as it is just a check pass.

mlir/test/tools/qcc-opt/jasp-check-static-qubit-allocation-test.mlir
```diff
@@ -0,0 +1,74 @@
...
+//       qc.x %3 : !qc.qubit
+//     } : !qc.qubit
+//     %4 = qc.measure %1 : !qc.qubit -> i1
+//     %5 = qc.measure %2 : !qc.qubit -> i1
+//     %6 = qc.measure %3 : !qc.qubit -> i1
+//     aux.record_bool %4
+//     aux.record_bool %5
+//     aux.record_bool %6
+//     return
+//   }
```
rainij: This is not good practice it should be its own test. Please add a new file for the cases were a failure occurs. Should look something like this

```mlir
// RUN: qcc-opt %s --check-static-qubit-allocation --split-input-file --verify-diagnostics

func.func @dynamic_size(%arg0: i64) {
    %0 = arith.index_cast %arg0 : i64 to index
    // expected-error @+1 {{qubit array allocation size must be a compile-time constant}}
    %alloc = memref.alloc(%0) : memref<?x!qc.qubit>
    return
}

// -----

func.func @zero_size() {
    // expected-error @+1 {{qubit array size must be positive, got 0}}
    %alloc = memref.alloc() : memref<0x!qc.qubit>
    return
}
```

The magic is done by the flags `--split-input-file --verify-diagnostics` and the separator `// -----` (related to the `--split-input-file` flag, could be overriden). 

I am not sure if the examples are as you intend them to be. Please check. In any case keep them as simple as possible. Your example looks a bit lengthy.

rainij: It might be a bit overkill to spend an extra file now for the case where no error occurs. Please check if it makes sense to add everything into one file (the various error cases and one case were no error occurs).

mlir/test/tools/qcc-opt/jasp-check-static-qubit-allocation-test.mlir
```diff
@@ -0,0 +1,74 @@
+// RUN: qcc-opt %s --check-static-qubit-allocation | FileCheck %s
+
+// Test that all the amount of memories we need to allocate for qubits are compile-time constants
+// The input is the result of a Canonicalization step
+func.func public @main() attributes {qcc.entry_point} {
```
rainij:
```suggestion
func.func public @main() {
```

The attributes play no role for your pass. Keep it simple.

mlir/test/tools/qcc-opt/jasp-check-static-qubit-allocation-test.mlir
```diff
@@ -0,0 +1,74 @@
...
+    %2 = memref.load %alloc[%c2] : memref<3x!qc.qubit>
+    qc.ctrl(%1) {
+      qc.x %2 : !qc.qubit
+    } : !qc.qubit
+    %3 = qc.measure %0 : !qc.qubit -> i1
+    %4 = qc.measure %1 : !qc.qubit -> i1
+    %5 = qc.measure %2 : !qc.qubit -> i1
+    aux.record_bool %3
+    aux.record_bool %4
+    aux.record_bool %5
```
rainij: Please avoid `aux.record`. We do not want to spread it through the code base without need. They are not essential for this test.

mlir/test/tools/qcc-opt/jasp-convert-memref-to-static-qubits-test.mlir
```diff
@@ -0,0 +1,56 @@
+// RUN: qcc-opt %s --staticize-qubit-refs | FileCheck %s
+
+// Test that the constant size `memref.alloc` are successfully converted to `qc.static` calls.
+// `memref.load` are substituted with direct accesses to the `qc.static` results.
+// The input is the result of a Canonicalization step which has successfully passed the `--check-static-qubit-allocation` test
+// After this stage, the IR is still in a intermediate step. It needs to be standardized using a Canonicalize step that remove all the `memref` dialect occurrences.
```
rainij: While I think it makes sense in general to document the tests (optionally) I feel you repeat alot of the docstring of the pass itself. Could you reduce this to an absolute minimum?

mlir/test/tools/qcc-opt/jasp-convert-memref-to-static-qubits-test.mlir
```diff
@@ -0,0 +1,56 @@
...
+    %2 = memref.load %alloc[%c2] : memref<3x!qc.qubit>
+    qc.ctrl(%1) {
+      qc.x %2 : !qc.qubit
+    } : !qc.qubit
+    %3 = qc.measure %0 : !qc.qubit -> i1
+    %4 = qc.measure %1 : !qc.qubit -> i1
+    %5 = qc.measure %2 : !qc.qubit -> i1
+    aux.record_bool %3
+    aux.record_bool %4
+    aux.record_bool %5
```
rainij: Remove those `aux.record_bool` ops.

mlir/test/tools/qcc-opt/jasp-convert-memref-to-static-qubits-test.mlir
```diff
@@ -0,0 +1,56 @@
...
+// CHECK: module {
+// CHECK-LABEL:   func.func public @main()
+// CHECK-SAME:                             attributes {qcc.entry_point} {
+// CHECK:     %[[C2:.*]] = arith.constant 2 : index
+// CHECK:     %[[C1:.*]] = arith.constant 1 : index
+// CHECK:     %[[C0:.*]] = arith.constant 0 : index
+// CHECK:     %[[Q0:.*]] = qc.static 0 : !qc.qubit
+// CHECK:     %[[Q1:.*]] = qc.static 1 : !qc.qubit
+// CHECK:     %[[Q2:.*]] = qc.static 2 : !qc.qubit
+// CHECK:     %[[ALLOC:.*]] = memref.alloc() : memref<3x!qc.qubit>
```
rainij: I notice that some `memref` ops are already cleared from the output. I see no clear pattern here. Some are gone, one is still left. Can you explain this and make more predictable?

mlir/test/tools/qcc-opt/jasp-staticize-qubit-refs-test.mlir
```diff
@@ -0,0 +1,56 @@
+// RUN: qcc-opt %s --staticize-qubit-refs | FileCheck %s
+
+// Test that the constant size `memref.alloc` are successfully converted to `qc.static` calls.
+// `memref.load` are substituted with direct accesses to the `qc.static` results.
+// The input is the result of a Canonicalization step which has successfully passed the `--check-static-qubit-allocation` test
+// After this stage, the IR is still in a intermediate step. It needs to be standardized using a Canonicalize step that remove all the `memref` dialect occurrences.
+func.func public @main() attributes {qcc.entry_point} {
```
rainij:
```suggestion
func.func public @test() {
```

mlir/test/tools/qcc-opt/jasp-staticize-qubit-refs-test.mlir
```diff
@@ -0,0 +1,56 @@
...
+    %4 = qc.measure %1 : !qc.qubit -> i1
+    %5 = qc.measure %2 : !qc.qubit -> i1
+    aux.record_bool %3
+    aux.record_bool %4
+    aux.record_bool %5
+    return
+  }
+
+// CHECK: module {
+// CHECK-LABEL:   func.func public @main()
```
rainij:
```suggestion
// CHECK-LABEL:   func.func public @test()
```

mlir/test/tools/qcc/ghz_from_qrisp.mlir
rainij: Hm OK. I see that the pipeline relies on it later on (at least the QIR export - otherwise the program never "outputs" anything as to what QIR is concerned). But of course we need a reliable way to insert those.

For now I suggest:
- Leave it here but document how exactly this IR was created and how you modified it.
- Leave this thread open, I will probably open an issue on adding a pass to create those `aux.record_bool` thingies in some fashion. ATM I am not sure how ugly of a solution this becomes or if it is a solution at all.

mlir/test/tools/qcc/ghz_from_qrisp_v2.mlir
rainij: I still see both. Please remove one of them and remove the suffix in the name.

rainij (changes requested):

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,10 +1,49 @@
...
+def CheckStaticQubitAllocation : Pass<"check-static-qubit-allocation"> {
+  let summary = "Validate that qubit array allocations use compile-time constant sizes.";
+  let description = [{
+      Verifies all qubit memref allocations have static sizes and reports diagnostic
+      errors for non-constant or invalid sizes.
+
+      An example of a correct IR is the following:
+      `%0 = "arith.constant"() <{value = dense<3> : tensor<i64>}> : () -> tensor<i64>
+       %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)`
+      in which the size of `jasp.create_qubit` is known at runtime.
```
rainij, line 25:
```suggestion
      An example of a correct IR is the following:

      ```mlir
      %0 = "arith.constant"() <{value = dense<3> : tensor<i64>}> : () -> tensor<i64>
       %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)
      ```

      in which the size of `jasp.create_qubit` is known at runtime.
```

rainij: Same for the other code snippets.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: As far as I am concerned, I never saw negative values to be used as a command for 'dynamic memory' (in languages like C where you use `malloc` family of functions. So for this commit, I left the documentation as it is. But if you see something different being the best way to go, I will gladly make the changes.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: Added all the modifications. I am sorry for the typos, my clang-tidy file was basically applying styling rules to plain text. Now I have fixed it.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: I think that if QRISP is our entry point, we should always point to their last version. Hence, if they make any change, it should be our concern to right away adapt to them. Otherwise we risk QRISP to move away from us and maybe people can find it as an obstacle in using FullStaQD. What is your opinion on that?

mlir/lib/Conversion/JaspToQC/JaspCheckStaticQubitAllocation.cpp
RiccardoRomanello: Yes, I agree with Reinhard actually. I think the explicit if statement is more readable and also less _formatting_ heavy by means of brackets.

mlir/lib/Conversion/JaspToQC/StaticizeQubitRefs.cpp
RiccardoRomanello: Thanks for the great suggestion, I changed it as you proposed.

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
RiccardoRomanello: Done

mlir/lib/Conversion/JaspToQC/CheckStaticQubitAllocation.cpp
RiccardoRomanello: Removed

mlir/test/tools/qcc-opt/jasp-convert-memref-to-static-qubits-test.mlir
RiccardoRomanello: I fixed the step and made everything more predictable. Looking at the step description in the .td file, you should be able to get the full internals of the step. Roughly speaking, right after the step, `qc` and `memref` co-exist as long as a canonicalization is not applied. The `qc` components are actually building up the program while `memref`s are just dangling, but at least predictable.

mlir/test/tools/qcc-opt/jasp-staticize-qubit-refs-test.mlir
RiccardoRomanello: Done

mlir/test/tools/qcc-opt/jasp-staticize-qubit-refs-test.mlir
RiccardoRomanello: Done

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: I applied the changes that you suggested. I forgot to use the markdown syntax.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
EnriqueNB: I think there is always a risk to point to an external source. It is best to reflect in the documentation what we have and not to point to an external source that we don't control. If they change something then it is our responsibility to update the code and the documentation.

mlir/lib/Conversion/JaspToQC/JaspCheckStaticQubitAllocation.cpp
EnriqueNB: I had to look at what code golf is :). I added the top if statement to not store the Walkresult result in a variable just for checking if it was interrupted. For me my version is most compact and readable, first you check the conditions for advancing in the Walk of the ops and then you interrupt. But it is a small change nonetheless, and readability is also subjective.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.h
```diff
@@ -4,6 +4,8 @@
 
 namespace qcc {
 #define GEN_PASS_DECL_JASPTOQC
+#define GEN_PASS_DECL_CHECKSTATICQUBITALLOCATION
+#define GEN_PASS_DECL_CONVERTMEMREFTOSTATICQUBITS
```
rainij:
```suggestion
#define GEN_PASS_DECL
```

This shortcut should should imply the three lines above. Compare also with the ToQIR passes where this is also done.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,17 +1,66 @@
 include "mlir/Pass/PassBase.td"
 
+def CheckStaticQubitAllocation : Pass<"check-static-qubit-allocation"> {
```
rainij: The pass name should probably refer to jasp in particular. For example: `jasp-check-static-qubit-allocation`.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
rainij: Your changes miss one tiny detail. For readability insert some empty lines. Like so:

```
      Text, followed by empty line ...

      ```mlir
      code
      ```

      text preceeded by empty line.
```

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,17 +1,66 @@
...
+      transforms `memref.load` on qubit arrays into direct `qc.static` references. Requires compile-time constant indices
+      (should be called after a `check-static-qubit-allocation` step). At the end of the step the `qc` and the `memref` dialect live together in the same IR.
+      The latter is actually dangling and it should be cleaned up via further `mlir::Canonicalization`.
+
+      Providing as input the following IR:
+      ```mlir
+      %0 = "arith.constant"() <{value = dense<3> : tensor<i64>}> : () -> tensor<i64>
+      %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)
+      ```
+      the pass will translate it into:
```
rainij, line 70: Here also some missing empty lines. Good for readability. See my other comment on the other pass.

mlir/lib/Conversion/JaspToQC/JaspCheckStaticQubitAllocation.cpp
rainij: @RiccardoRomanello please try to apply this suggestion (you can click the button above).

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
rainij: @RiccardoRomanello please try to apply this suggestion.

mlir/test/tools/qcc-opt/jasp-check-static-qubit-allocation-test.mlir
rainij: @RiccardoRomanello

mlir/test/tools/qcc/ghz_from_qrisp.mlir
rainij: At least the step with the stablehlo stuff should not be needed. Related to this issue https://github.com/FullStaQD/compiler/issues/48 . For the `aux.record_bool` I have already created this issue: https://github.com/FullStaQD/compiler/issues/50 . For the entrypoint thing there is no issue, but it is good that you have documented this here.

In the near future all these manual transformations should not be needed at all. We should be able to take the qrisp output directly. So please add a TODO somehwere in front of the list.

rainij (changes requested):
@RiccardoRomanello you did not answer to some of my threads from the last review. Please have a look, you probably just missed them.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.h
RiccardoRomanello: Done, it works. Thanks for the suggestion, I like it.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: Changed the name and related calls.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: Done in both the files

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: @rainij what do you think about it?

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: Done

mlir/lib/Conversion/JaspToQC/JaspCheckStaticQubitAllocation.cpp
RiccardoRomanello: Fixed. I tried to follow the yellow underlined stuff given by VSCode but it is actually too strong.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
RiccardoRomanello: Done here as well. Sorry but I don't like too much the 'apply suggestion' of the Web UI if it is on code... I prefer to test It locally.

mlir/test/tools/qcc-opt/jasp-check-static-qubit-allocation-test.mlir
RiccardoRomanello: Fixed with one single file

mlir/test/tools/qcc-opt/jasp-convert-memref-to-static-qubits-test.mlir
RiccardoRomanello: Reduced to the minimum, trying to avoid explaining things again.

mlir/test/tools/qcc/ghz_from_qrisp.mlir
RiccardoRomanello: Done, added the TODO

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -1,17 +1,74 @@
...
+      %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)
+      ```
+
+      in which the size of `jasp.create_qubit` is known at runtime.
+
+      On the other hand, in the following snippet of code:
+
+      ```mlir
+      func.func public @main(%arg14: !jasp.QuantumState, %s: tensor<i64>) attributes { qcc.entry_point } {
+        %1, %2 = "jasp.create_qubits"(%s, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)
```
rainij, line 31:
```suggestion
      func.func public @main(%arg14: !jasp.QuantumState, %s: tensor<i64>) attributes { qcc.entry_point } {
        %1, %2 = "jasp.create_qubits"(%s, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)
        ...
      }
```

rainij: Code snippets should ideally have valid syntax. I suggest to use `...` to explicitly omit stuff. IMO closing parens always also makes it more readable.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
rainij: That is totally fine. It is just the easiest way for me to communicate what I want.

mlir/lib/Conversion/JaspToQC/JaspCheckStaticQubitAllocation.cpp
rainij: Yes concerning the "unused header warning": That is a typical problem in MLIR, they are only implicitly needed by things like

```cpp
#define GEN_PASS_DEF_JASPCHECKSTATICQUBITALLOCATION
#include "qcc/Conversion/JaspToQC/JaspToQC.h.inc"
```

The linter cannot analyze this situation hence the false positive. You can use this to eliminate the false positives (not perfect but better then nothing):

https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md

Lukas pointed it out some time ago.

mlir/test/tools/qcc-opt/jasp-check-static-qubit-allocation-test.mlir
rainij: 👍

rainij (changes requested):
Just a few tiny things left.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: I fixed the code. Unfortunately GitHub was not showing me comment in the file tab, so I had to do an external commit.

rainij (approved):

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
rainij: I have seen this was done.

mlir/test/tools/qcc/ghz_from_qrisp_v2.mlir
rainij: I see that they are gone. Good.

rainij:
@burgholzer this can be merged from my side.

@EnriqueNB please have a look if your requests have been addressed.

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -27,7 +27,12 @@ void buildQuantumPipeline(mlir::PassManager& pm) {
   // whenever possible.
   pm.addPass(qcc::createJaspToQC());
 
-  // Initial cleanup
+  // dynamic to static allocation translation
```
burgholzer:
```suggestion
  // Dynamic to static allocation translation
```
consistency.

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -27,7 +27,12 @@ void buildQuantumPipeline(mlir::PassManager& pm) {
...
+  pm.addPass(mlir::createCanonicalizerPass());
+  pm.addPass(qcc::createJaspCheckStaticQubitAllocation());
+  pm.addPass(qcc::createConvertMemrefToStaticQubits());
+
+  // cleanup
```
burgholzer:
```suggestion
  // Cleanup
```
consistency.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
```diff
@@ -0,0 +1,134 @@
...
+      // Retrieve the pre-allocated physical qubit from our map.
+      auto it = qubitMap.find({baseMemref, idx});
+      if (it == qubitMap.end()) {
+        loadOp->emitError("internal error: static qubit not found for this allocation");
+        return WalkResult::interrupt();
+      }
+
+      // Replace all uses of the 'loaded' qubit with the 'static' hardware qubit.
+      loadOp.getResult().replaceAllUsesWith(it->second);
+      // loadOp.erase();
```
burgholzer: This is commented-out code. It should either be removed or uncommented.
I'd be in favor of explicitly deleting the operation since we already have a handle on the operation and know that it shall be removed.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+      return WalkResult::advance();
+    });
+
+    if (secondResult.wasInterrupted()) {
+      return signalPassFailure();
+    }
```
burgholzer, line 129: I would argue there should be a third walk of the memref alloc operations that removes them since they are no longer needed. This should be much higher performance than relying on canonicalization to pick up the operations.
We already know that the operations are no longer used since we replaced all their uses; so we might as well directly remove the operation itself.
Note that this will also require a docstring update.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
```diff
@@ -0,0 +1,134 @@
...
+        return WalkResult::interrupt();
+      }
+
+      int64_t size = memrefType.getDimSize(0);
+      builder.setInsertionPoint(allocOp);
+
+      // Create a unique 'static' op for every slot in the memref.
+      for (int64_t i = 0; i < size; i++) {
+        auto staticQubit = qc::StaticOp::create(builder, allocOp->getLoc(), qc::QubitType::get(op->getContext()),
+                                                nextGlobalQubitIdx++);
```
burgholzer:
```suggestion
        auto staticQubit = qc::StaticOp::create(builder, allocOp->getLoc(), 
                                                nextGlobalQubitIdx++);
```
Fairly sure this is not needed.

burgholzer (changes requested):
Overall, this LGTM. Just two very small consistency comments and one slightly larger thought about leaving cleanup to canonicalization, which I believe is not necessarily a good idea for performance.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
rainij: I definitely know that in general one can and should leave _some_ cleanup (including removing unused stuff) to canonicalisation and similar passes. But there are cases were one recommends against it, maybe non-pure stuff like `memref.alloc` is one of these. 

Just wanted to mention this. Otherwise I tend to believe that you are right in this particular case and it is better to do it ourselves (here).

EnriqueNB (approved):
Good from my side.

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
burgholzer: Yeah, fully agree that some cleanup can/should be left to canonicalization. 
In this case here, the cleanup requires a sequence of resolutions though. 
Since `load` operations are also still part of the program, canonicalization first needs to remove all load operations before it can remove the memref.
Especially for cases where you already have a handle on the operation to be erased, it is a waste to not eliminate it directly. As far as I understand, canonicalization repeatedly iterates through the program trying to match patterns. The smaller the program, the more efficient the iteration.

RiccardoRomanello: Thanks for the suggestion!
I successfully translated the call to the canonicalizer with a third walk that removes any memref reference. 
Updated the td, test files and potential comments to avoid mentioning 'need canonicalization'.

burgholzer (approved):
LGTM now!

mlir/lib/Conversion/JaspToQC/ConvertMemrefToStaticQubits.cpp
rainij: Seems to be resolved.

rainij:
The github UI tells me that there are still unresolved threads. I am absolutely not able to find those. No clue. Maybe it is about merging main, lets see if if this works.

Update: didn't work. I merge anyway.

PR #44, opened 2026-05-11
Add license file

### LINKED ISSUE #42: Add license file to repo
Add Apache2.0 license to the repo and enforce a small license statement inside of each file via a pre-commit hook.

denialhaag:
Fixes #42


denialhaag:
@burgholzer and @rainij, this is a first proposal.

As requested in #42, we are using `//` instead of `/* ... */`. I have also configured the tool to add the filename.

In QDMI, the license header is more verbose; we could do the same here.

I'm happy to further tweak this. 🙂

burgholzer (approved):
I am fine with this as-is. Thanks for setting this up.
I'll leave it to @rainij for the final verdict.

LICENSE
```diff
@@ -0,0 +1,278 @@
+==============================================================================
+The LLVM Project is under the Apache License v2.0 with LLVM Exceptions:
```
rainij: Shouldn't we mention our project instead here?

mlir/cmake/ExternalDependencies.cmake
```diff
@@ -1,3 +1,10 @@
+# ExternalDependencies.cmake
```
rainij: Is it overkill to give cmake files a license header too? It looks like LLVM does not do this.

mlir/include/qcc/Compiler/Pipeline.h
```diff
@@ -1,3 +1,10 @@
+// Pipeline.h
+//
+// Copyright (c) 2026 FullStaQD Project
+// All rights reserved.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
rainij, line 6: I would actually prefer to follow closely the LLVM style. Here is an example:

```cpp
//===---- RISCVISelDAGToDAG.h - A dag to dag inst selector for RISC-V -----===//
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
//
// This file defines an instruction selector for the RISC-V target.
//
//===----------------------------------------------------------------------===//
```

It catches the eye much better IMO. I think in your PR I would not give you the burden of filling out the details but you could leave TODO lines, like this:

```cpp
//===---- Pipeline.h - TODO: summary of file -----------------------------===//
//
// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM Exceptions.
// See <repo-root>/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
//
// TODO: file docstring
//
//===----------------------------------------------------------------------===//
```

I do not know what exactly the CI checks, but IMO a heuristic to check it is there is enough.

rainij: Also, I am not 100% about the exact text. The current text looks a bit minimalistic. I would like somebody else to have a look (@MatthiasTraube ?).

It looks like a small detail but I would like to get this right. Otherwise I fear regrets in the future.

rainij (changes requested):
Thanks for taking care. I left a few comments. I would like to iterate on this to get it right. IMO it is an important aspect of the project to get right.

rainij:
@MatthiasTraube I also requested your review. Maybe you can look at it from your angle. In particular I would like to get the header text right. I believe staying as close as possible to what LLVM does might be a good option for us as well.

LICENSE
denialhaag: Thanks for catching that! Done.

mlir/cmake/ExternalDependencies.cmake
denialhaag: I have removed the headers from all CMake files. 👍

mlir/include/qcc/Compiler/Pipeline.h
denialhaag: If we really want this, I don't think we can use [`license-tools`](https://github.com/emzeat/mz-lictools). Honestly, I don't even see a way to use any kind of automated pre-commit hook to _write_ these kinds of license headers. At most, we could write a workflow that checks the presence and fails if something is off.

Edit: I'm only speaking about the summaries. The "design" and text can easily be adapted.

That said, I personally don't like the idea of mixing license headers with docstrings. The summary line at the top of the file seems like a prime candidate for becoming outdated as files grow. That's just my opinion, though.

burgholzer: ~~I am not sure whether the tool we use to automatically generate these can handle these per-file variations.~~ Daniel already covered that part

It surely can handle something along the lines of
```
//===---- Pipeline.h ------------------------------------------------------===//
//
// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM Exceptions.
// See <repo-root>/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
```

And I mostly agree with Daniel, that mixing documentation and license information feels a bit off.
The above suggestion is automatable and concise.

denialhaag: I have looked into the source code of `license-tools` a bit more now, and it does not appear to support any templating. The filename appears to be accessible only from the `title` field. 😕

burgholzer: 😢 What's the best we could do with that?

denialhaag: Something like this, with a variable number of empty lines. 😬 

```
// Pipeline.h
//
//===----------------------------------------------------------------------===//
//
// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM Exceptions.
// See <repo-root>/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
```

burgholzer: Hm. I'd personally just rip out the filename here completely and just stick with
```
//===----------------------------------------------------------------------===//
//
// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM Exceptions.
// See <repo-root>/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
```

I do not see any particular value in adding the filename itself to the license header.

denialhaag: I also wouldn't mind removing the filename. 🤔

MatthiasTraube: I think the format 

```
//===----------------------------------------------------------------------===//
//
// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM Exceptions.
// See <repo-root>/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
```

is completely fine. There is no need to give a summary of the file, nor to give the filename in the auto generated header.

MatthiasTraube (approved):

mlir/include/qcc/Compiler/Pipeline.h
rainij: I agree with @MatthiasTraube, just two details: I would like to standardize the _optional_ file docstring a bit. IMO the way LLVM solved it is really neat.

So I would propose this format if there is not file-level docstring:

```cpp
//===----------------------------------------------------------------------===//
//
// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM Exceptions.
// See <repo-root>/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
```

And this one if there is a docstring:

```cpp
//===----------------------------------------------------------------------===//
//
// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM Exceptions.
// See <repo-root>/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
//
// Docstring of the file would go here.
//
//===----------------------------------------------------------------------===//
```

The alternative would be two such block comments separated by some newlines. IMO this would look ugly as hell. I just mention it because I want to make sure the tooling does not block the second format. If everybody agrees I would like to try out this format in at least one of the files to check that it works.

The second small issue: The first sentence "Part of ... Exceptions." is slightly wider than the header itself (which is 80 characters). A simple solution would be to use e.g. 90 characters (or 100). (Elsewhere we have 120 chars in some formatter, but I have to say I am not 100% happy with 120 chars as this makes two windows side by side on the same monitor difficult even on a big screen) Another would be to actively break the line before the word "Exception".

@burgholzer @MatthiasTraube:

Do you agree with the optional file docstring?

What is your opinion on the second issue (line width)?

denialhaag: `license-tools` automatically wraps lines longer than 80 characters. The best we can do with the tool:

```
// Copyright (c) 2026 FullStaQD Project
// All rights reserved.
//
// ===----------------------------------------------------------------------===//
//
// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM
// Exceptions.
// See <repo-root>/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
// ===----------------------------------------------------------------------===//
```

I also just realized that there is no way to get rid of the first two lines. At this point, we might have to rethink whether we want to use `license-tools` or look for something else. I guess the tool has a very limited scope, focusing on creating SPDX-compliant license headers.

rainij: Yes I would then recommend to switch to a tool which supports the format we actually want. I would not want a tool to dictate us how it should look. On the other hand it is OK if the CI only has a heuristic check that the license header is there. IMO it is not critical to be able to auto-generate it (but nice to have of course).

On the other hand I really do not know what this copyright notice implies. Many projects have both: license + plus copyright. LLVM seems to only mention this license in its header. Might make sense to investigate the implications.

rainij: So I googled a bit around: It _seems_ the copyright notice (and especially the "All rights reserved" part) is a legacy thing. Gemini mentioned that according to the Berne Convention (yes [this one](https://en.wikipedia.org/wiki/Berne_Convention) from 1886) the copyright notice is redundant:

> The Berne Convention introduced the concept that protection exists the moment a work is "fixed", that is, written or recorded on some physical medium, and its author is automatically entitled to all copyrights in the work and to any [derivative works](https://en.wikipedia.org/wiki/Derivative_work), unless and until the author explicitly disclaims them or until the copyright expires. A creator need not [register](https://en.wikipedia.org/wiki/Copyright_registration) or "apply for" a copyright in countries adhering to the convention. It also enforces a requirement that countries recognize rights held by the citizens of all other parties to the convention. Foreign authors are given the same rights and privileges to copyrighted material as domestic authors in any country that ratified the convention. The countries to which the convention applies created a Union for the protection of the rights of authors in their literary and artistic works, known as the Berne Union.

One might wonder why one then finds copyright notices everywhere. I believe the reason is that the USA only joined in 1989 and that the habit of adding it dies hard (in the sense of safety first, or just pure ignorance "we did it always like this").

Gemini also claims that "All rights reserved" actually contradicts the Apache lincense. The gist is:

- "All rights reserved": You have no permission to copy this work.
- Apache: You have the right to copy this work.

Sounds somewhat plausible to me - I did not check though. I would take LLVM avoiding this copyright thingy entirely as the primary motivation to leave it out for us too.

rainij: And btw: 80 chars is OK from my side. I think it looks totally fine with the line break.

denialhaag: Alright, I have switched to `license-eye` now. If we want, the action can also be set up to fix the headers: https://github.com/apache/skywalking-eyes#fix-license-headers

burgholzer (changes requested):
This looks reasonable to me now.
I think this will need some form of documentation to show developers how to properly maintain the license headers.
Can this be run via prek?

denialhaag:
> I think this will need some form of documentation to show developers how to properly maintain the license headers.

Good point. Done.

> Can this be run via prek?

There is no first-party support. We could define it as a system hook, but I'm not sure we want to, since it would require every developer to install `license-eye`.

mlir/include/qcc/Compiler/Pipeline.h
rainij: The format looks good to me now.

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir.mlir
rainij, line 1: If we follow the LLVM project then `*.mlir` and cmake files should not have these license headers. I am also in favor of not having them there as they introduce clutter. But honestly I really do not know when they are needed and why LLVM omits them except for the "real" source files.

From my googling earlier today I also found that these SPDX identifiers seem important to lawyers. It might be an option to just use those for mlir and cmake files:

```
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

But I would only do this if there is a true reason for it.

@burgholzer @MatthiasTraube sorry to bother again but I want to make sure everybody is on the same page. What is you oppinion? Do we have any expert at our disposal who can answer such questions?

LICENSE.txt
rainij, line 1: IMO the filename should be `LICENSE.md` or `LICENSE.txt` (in case it is not compatible with markdown).

LICENSE
```diff
@@ -0,0 +1,278 @@
...
+terms. All such code will be identified clearly using at least one of two
+mechanisms:
+1) It will be in a separate directory tree with its own `LICENSE.txt` or
+   `LICENSE` file at the top containing the specific license and restrictions
+   which apply to that software, or
+2) It will contain specific license and restriction terms at the top of every
+   file.
+
+==============================================================================
+Legacy LLVM License (https://llvm.org/docs/DeveloperPolicy.html#legacy):
```
rainij: This section seems odd. I believe it _must_ be removed. We should consider tayloring the license file to our needs.

LICENSE
```diff
@@ -0,0 +1,278 @@
...
+Software from third parties included in the LLVM Project:
+==============================================================================
+The LLVM Project contains third party software which is under different license
+terms. All such code will be identified clearly using at least one of two
+mechanisms:
+1) It will be in a separate directory tree with its own `LICENSE.txt` or
+   `LICENSE` file at the top containing the specific license and restrictions
+   which apply to that software, or
+2) It will contain specific license and restriction terms at the top of every
+   file.
```
rainij:
```suggestion
==============================================================================
Third-party software used by the FullStaQD Project:
==============================================================================
The FullStaQD Project is based on the LLVM Project. Software from the LLVM
Project is typically licensed under the Apache License v2.0 with LLVM 
Exceptions. 

For more information on LLVM's licensing, see:
https://llvm.org/docs/DeveloperPolicy.html#new-llvm-project-license-framework

Other third-party software will be identified clearly using at least one of 
two mechanisms:
1) It will be in a separate directory tree with its own LICENSE.txt file.
2) It will contain specific license and restriction terms at the top of 
   the file.
```

rainij: I generated this using gemini. Seems plausible to me. But again, I am no lawyer.

rainij: Also, I used the wording "FullStaQD" (we use it also elsewhere). I am not sure if it is a problem that this compiler is intended to be renamed into something else. But I believe one can change it later on.

rainij (changes requested):
Sorry for the additional requests but my feeling is that nobody of us really knows about licensing. So I gave it some additional thoughts.

LICENSE
denialhaag: I have applied your suggestion, but I'll leave this thread open for now so that others can comment.

LICENSE
denialhaag: Good catch. Done.

LICENSE.txt
denialhaag: I've renamed it to `LICENSE.txt` because it would look a bit weird as rendered Markdown.

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir.mlir
denialhaag: I have removed the headers from all `*.mlir` files now. I'd vote against using different headers for different file types because this will open another can of worms. I don't think `license-eye` could handle this.

MatthiasTraube: AI says that the SPDX is a tooling convenience, i.e. it MUST be the exact wording

```
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
since this will be recognized by automated tooling for checking license compatibility. The point seems to be that in bigger companies usually there isn't somebody checking license by hand, but a tool scanning the code base, which recognizes the SPDX identifier and signals green light for Apache2.0 (hopefully). The SPDX identifier system is apparently maintained by the Linux foundation.

Furthermore, AI tells me that the inconsistencies of using the small license header file in the LLVM repo are out of convenience or for historic reasons. The proper way seems to be to just have it in all files. However, for somewhat generic files like the cmake files or very small files it seems a bit like overkill and clutters the code base. I don't have a strong opinion on this, though, and leave the decision to @rainij and @burgholzer whether to include the SPDX identifier just everywhere.

Finally the copyright: AI suggests that we can use a NOTICE.txt file with a small copyright notice:

```
NOTICE

FullStaQD Project
Copyright (c) 2026 FullStaQD Contributors

This product includes software developed as part of the FullStaQD Project.
```
This serves a different purpose. The license stays in place and is not touched. Adding a NOTICE.text file with the above text ensures, that the original contributors, i.e. the FullStaQD project, are acknowledged by everybody using the code base, since the NOTICE.txt file must be carried along when redistributing.

Thus, I suggest to just include the NOTICE.txt in the root folder and add the header we now agreed on in every file it doesn't clutter the code.

rainij: :+1: for the `NOTICE.txt` .

:+1: for making license header optional in cmake and mlir files (not checked by tooling). I suggest that the default is to not include it.

It seems it is always easy to just add a license header to a file which did not have one before. No header is often interpreted as "All rights reserved" by default (says gemini, but I heard this along time ago elsewhere). "Often" the top-level `LICENSE.txt` is even enough so that _any_ license header (even in C++ files) is actually considered redundant. Gemini claims people are just extremely conservative when it comes to those things.

In my previous company we also worked with automatic license checkers. At least for the big car manufacturers I can confirm that they are paranoid with licenses. But for a reason actually - there really were black sheep in the open source community in the past: https://en.wikipedia.org/wiki/Open_source_license_litigation . If one looks at the dates mentioned in the wikipedia article the problem seems to be not so relevant anymore, but the measures against it probably stay in place for some time longer.

burgholzer: I think we are in a space where we do not need to be overly cautious here.
My impression was always that the main license file is the most important thing and that the SPDX classifier helps automated tooling. Everything else is pretty much optional/conservative.
So I believe we should simply find a pragmatic solution, which it seems we have found now. 😌 
We are following a permissive open-source license anyway. I guess this is the simplest case compared to proprietary licenses or copyleft licenses.

MatthiasTraube: I agree with Lukas, let's not over complicate now, since nobody of us is a lawyer and we are following a standard pattern for open source projects.

denialhaag: I have added a `NOTICE.txt` file now. 👍

denialhaag:
From my side, this is ready to go in now. 😌

burgholzer (approved):
I am fine with this too

LICENSE
rainij: Thanks, seems nobody wants to change the wording, so I resolve the thread.

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir.mlir
rainij: Cool, I feel actually much better now having this discusse through :slightly_smiling_face:

rainij (approved):

PR #53, opened 2026-05-20
Document qrisp mlir

### LINKED ISSUE #48: Document Qrisp MLIR generation
I Qrisp "frontend" assumes that the input IR we receive went roughly through the following process:

- Generate mlir via `Jaspr.to_mlir()` (contains stablehlo)
- Lower stablehlo to standard dialects - somehow.

An example for the first step is attached below. However, we never documented it and especially have no standard way for the second step. Options are:

- Use something like the script provided by Raphael Seidel: See https://github.com/FullStaQD/compiler/pull/18#issuecomment-4209658674
- Build [stablehlo](https://github.com/openxla/stablehlo) and use `stablehlo-opt` (only if first option is not going to work)

We hope that this simplifies once https://github.com/eclipse-qrisp/Qrisp/pull/528 hits the released version of Qrisp. But in the mean-time we need something that works. And in any case the first part needs some documenting.

# The task

I propose to add a subfolder `mlir/test/tools/qcc/qrisp/` were we put all tests starting from qrisp as "frontend". This folder should get a `README.md` documenting what needs to be done to produce MLIR from a given python program. Each test should always document the corresponding python script from which it was derived.

# Appendix: Using `to_mlir()`

```python
from qrisp import *

def example():
    qv = QuantumVariable(2)
    h(qv[0])
    cx(qv[0], qv[1])
    return measure(qv)

jaspr = make_jaspr(example)()

mlir_stablehlo = jaspr.to_mlir()

print(mlir_stablehlo)
```

RiccardoRomanello:
Branch to document the generation of MLIR starting from Qrisp input. 
Related to #48


rainij (changes requested):
Can you please cleanup this PR before review it seems to contain your other PR.

PR #59, opened 2026-05-26
Allow one-line blocks without braces

### LINKED ISSUE #58: Change formatters to allow unbraced then-blocks
For consistency with LLVM coding styles, we should allow 
```c
if (condition)
    single_line_body();
```

julfarn:
Adresses #58

### DIFF
--- a/mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
+++ b/mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
@@ -74,9 +74,8 @@ public:
     });
 
     addConversion([](mlir::RankedTensorType type) -> std::optional<mlir::Type> {
-      if (type.getRank() == 0) {
+      if (type.getRank() == 0)
         return type.getElementType();
-      }
       return std::nullopt; // Leave multi-dimensional tensors alone
     });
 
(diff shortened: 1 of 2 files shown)


rainij (approved):

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -74,9 +74,8 @@ class JaspToQCTypeConverter : public TypeConverter {
     });
 
     addConversion([](mlir::RankedTensorType type) -> std::optional<mlir::Type> {
-      if (type.getRank() == 0) {
+      if (type.getRank() == 0)
```
burgholzer, line 77: I feel this is a good example of the inconsistencies that I pointed out in the related issue.
Now this PR changed this single line. But in the same file there are at least two more simple one-liner `if` statements that still have braces.
As far as aesthetics go, in my opinion, this is worse than a couple extra braces.

burgholzer (changes requested):
Just to make explicit what I meant in the issue.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij: This could be resolved by a follow up PR, but not important anymore as we do not merge.

rainij:
We (@julfarn , @burgholzer) discussed the tradeoffs and came to the conclusion that there are good arguments for and against it. Based on our targeted user base staying with a more strict rule might make sense. Hence we close this PR.

PR #62, opened 2026-05-27
Simplified default dev container

rainij:
Since it is possible to have a totally custom dev container config under `.devcontainer/local/` (documented in README) we do not have to bloat our default devcontainer with extra features.


rainij:
FYI: @julfarn @RiccardoRomanello @EnriqueNB (just in case you use this feature which will get deleted by this PR)

julfarn:
Not sure how much I like this. This will force me to use a custom dev container. If everyone has a custom one, it's easy to get out of sync. What even is the purpose of having a shared one then?

rainij:
> Not sure how much I like this. This will force me to use a custom dev container. If everyone has a custom one, it's easy to get out of sync. What even is the purpose of having a shared one then?

@julfarn it was never about a shared devcontainer. At least this was not my intention. This devcontainer you see under vc is just for quick onboarding. I always expected that after some time people naturally feel constrained and the urge to make their own. At least I felt the urge. But even then it naturally serves as the template for your own custom devcontainer.

The alternative would be to clutter the "shared" devcontainer with all kinds of tweaks. For example I also need something related to Claude in it (and it is very peculiar). I think it would not great to put this into vc.

burgholzer (approved):
I am in favor. Keeps the common case simpler and feels like a more natural starting point.

PR #65, opened 2026-06-01
Adding qcc entrypoint attribute to input mlir

### LINKED ISSUE #60: Pass to add `qcc.entrypoint` attribute
The back end part of the pipeline relies on the `qcc.entrypoint` attribute to know where to start the program.  Our current quasi-frontend Qrisp has the convention to call the entrypoint `@main` in the generated MLIR. Create a pass (e.g. named `add-entrypoint-to-main`) which adds exactly this attribute to this function.

IMO it could even make sense to allow for an option to specify the name of the entrypoint. But in the simplest case hardcoding `main` is probably ok.

Note that we have to see how this evolves in the future when the program input becomes more complex (multiple kernels, a classical entrypoint etc).

RiccardoRomanello:
Resolves #60

### DIFF
--- a/mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
+++ b/mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
@@ -62,16 +62,16 @@ def ConvertMemrefToStaticQubits : Pass<"convert-memref-to-static-qubits"> {
       Providing as input the following IR :
 
       ```mlir
-        %0 = "arith.constant"()<{value = dense<3>: tensor<i64>}> : ()->tensor<i64>
-        %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) ->(!jasp.QubitArray, !jasp.QuantumState)
+      %0 = "arith.constant"()<{value = dense<3>: tensor<i64>}> : () -> tensor<i64>
+      %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) ->(!jasp.QubitArray, !jasp.QuantumState)
       ```
 
       the pass will translate it into:
 
       ```mlir
-        %0 = qc.static 0 : !qc.qubit
-        %1 = qc.static 1 : !qc.qubit
-        %2 = qc.static 2 : !qc.qubit
+      %0 = qc.static 0 : !qc.qubit
+      %1 = qc.static 1 : !qc.qubit
+      %2 = qc.static 2 : !qc.qubit
       ```
 
       in which the allocation of three qubits has been turned into 3 separate static allocations.
--- a/mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
+++ b/mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
@@ -80,7 +80,7 @@ def ConvertMemrefToStaticQubits : Pass<"convert-memref-to-static-qubits"> {
 }
 
 def AddEntrypointToMain : Pass<"add-entrypoint-to-main", "mlir::ModuleOp"> {
-  let summary = "Add the `qcc.entry_point` attribute to the entry-point function.";
+  let summary = "Adds the `qcc.entry_point` attribute to any function called `@main`.";
   let description = [{
       The QIR conversion pipeline relies on the `qcc.entry_point` attribute to
       identify the function from which a quantum program starts. Frontends such
--- a/mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
+++ b/mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
@@ -56,27 +56,62 @@ let dependentDialects = ["mlir::qc::QCDialect", "mlir::tensor::TensorDialect", "
 
 def ConvertMemrefToStaticQubits : Pass<"convert-memref-to-static-qubits"> {
   let summary = "Replace dynamic qubit array loads with static qubits.";
-  let description = [{ First, substitutes `memref.alloc` with  a list of `qc.static` allocation. Subsequently,
-      transforms `memref.load` on qubit arrays into direct `qc.static` references. Requires compile-time constant indices
-      (should be called after a `check-static-qubit-allocation` step).
+  let description = [{ Fir st,
+      substitutes `memref.alloc` with a list of `qc.static` allocation.Subsequently,
+      transforms `memref.load` on qubit arrays into direct `qc.static` references.Requires compile -
+          time constant indices(should be called after a `check - static - qubit - allocation` step)
+                  .
 
-      Providing as input the following IR:
+              Providing as input the following IR :
 
-      ```mlir
-      %0 = "arith.constant"() <{value = dense<3> : tensor<i64>}> : () -> tensor<i64>
-      %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)
+      ```mlir %
+              0 = "arith.constant"()<{value = dense<3>: tensor<i64>}> : ()->tensor<i64> % 1,
+      % 2 = "jasp.create_qubits"(% 0, % arg14)
+      : (tensor<i64>, !jasp.QuantumState)
+            ->(!jasp.QubitArray, !jasp.QuantumState)
       ```
 
-      the pass will translate it into:
+        the pass will translate it into :
 
-      ```mlir
-      %0 = qc.static 0 : !qc.qubit
-      %1 = qc.static 1 : !qc.qubit
-      %2 = qc.static 2 : !qc.qubit
+      ```mlir %
+            0 = qc.static 0 : !qc.qubit % 1 = qc.static 1 : !qc.qubit % 2 = qc.static 2
+      : !qc.qubit
       ```
 
-      in which the allocation of three qubits has been turned into 3 separate static allocations.
-
+        in which the allocation of three qubits has been turned into 3 separate static allocations.
 }];
 let dependentDialects = ["mlir::qc::QCDialect"];
+}
+
+def AddEntrypointToMain : Pass<"add-entrypoint-to-main", "mlir::ModuleOp"> {
+  let summary = "Marks the entry-point function with the `qcc.entry_point` attribute.";
+  let description = [{
+      The QIR conversion pipeline relies on the `qcc.entry_point` attribute to
+      identify the function from which a quantum program starts.
+
+      This pass first checks whether *any* function in the module already
+      carries the `qcc.entry_point` attribute. If so, the pass does nothing --
+      the existing entry-point is respected, whether it is `@main` or any other
+      function annotated elsewhere.
+
+      Otherwise, the pass looks up the function whose name matches the
+      configured entry-point name (`main` by default) and attaches the
+      `qcc.entry_point` unit attribute to it. This targets the convention used
+      by frontends such as qrisp, which name the entry-point `@main` but do not
+      attach the attribute themselves.
+
+      The pass emits an error only if no entry-point exists *and* no function
+      with the configured name can be found.
+
+      TODO: This convention will likely have to evolve once the program input
+      becomes more complex (multiple kernels, a separate classical entry-point,
+      etc.).
+  }];
+
+  let options = [
+    Option<"entryPointName", "entry-point-name", "std::string", /*default=*/"\"main\"",
+           "Name of the function to be marked as the entry-point.">
+  ];
+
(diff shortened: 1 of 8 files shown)


mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -56,27 +56,62 @@ let dependentDialects = ["mlir::qc::QCDialect", "mlir::tensor::TensorDialect", "
...
-      %1 = qc.static 1 : !qc.qubit
-      %2 = qc.static 2 : !qc.qubit
+      ```mlir %
+            0 = qc.static 0 : !qc.qubit % 1 = qc.static 1 : !qc.qubit % 2 = qc.static 2
+      : !qc.qubit
       ```
 
-      in which the allocation of three qubits has been turned into 3 separate static allocations.
-
+        in which the allocation of three qubits has been turned into 3 separate static allocations.
```
rainij: Please revert the changes to `convert-memref-to-static-qubits`, looks like an accident.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -56,27 +56,62 @@ let dependentDialects = ["mlir::qc::QCDialect", "mlir::tensor::TensorDialect", "
...
 }];
 let dependentDialects = ["mlir::qc::QCDialect"];
 }
+
+def AddEntrypointToMain : Pass<"add-entrypoint-to-main", "mlir::ModuleOp"> {
+  let summary = "Marks the entry-point function with the `qcc.entry_point` attribute.";
```
```suggestion
  let summary = "Adds the `qcc.entry_point` attribute to any function called `@main`.";
```

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -56,27 +56,62 @@ let dependentDialects = ["mlir::qc::QCDialect", "mlir::tensor::TensorDialect", "
...
+  let description = [{
+      The QIR conversion pipeline relies on the `qcc.entry_point` attribute to
+      identify the function from which a quantum program starts.
+
+      This pass first checks whether *any* function in the module already
+      carries the `qcc.entry_point` attribute. If so, the pass does nothing --
+      the existing entry-point is respected, whether it is `@main` or any other
+      function annotated elsewhere.
```
rainij: Here I am not sure if we need this check. I believe QIR even allows multiple entrypoints, please check.

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -29,6 +29,9 @@ void buildQuantumPipeline(mlir::PassManager& pm) {
   // Qrisp output contains a lot of functions that can be trivially inlined.
   pm.addPass(mlir::createInlinerPass());
 
+  // Attach the `qcc.entry_point` attribute. Add it before anything else runs.
```
rainij: Comment probably not needed.

rainij:
> Relates to #60

@RiccardoRomanello please write "Resolves #60". Otherwise the the issue is not auto-closed. Only write "relates" if the PR does not fully close the issue.

rainij:
> First commit to add qcc entrypoint

@RiccardoRomanello this is a bad title for the PR.

rainij (changes requested):

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: Fixed, sorry for the issue.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: I just applied the marks to add change. Actually, with the option `--add-entrypoint-to-main="entry-point-name=NAME"` we can already support adding the attribute to a custom function. So I would not just go for main in the description of the pass. But I am happy to hear about your point of view on that.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: I checked this page on [QIR Base Profile](https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Base_Profile.md#attributes) and it actually always mention `entry point` with no plural, so always talking about function and not functions.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -49,34 +49,65 @@ def JaspToQC : Pass<"jasp-to-qc"> {
...
+  let description = [{ First, substitutes `memref.alloc` with a list of `qc.static` allocation. Subsequently,
+      transforms `memref.load` on qubit arrays into direct `qc.static` references.
+      Requires compile-time constant indices (should be called after a `jasp-check-static-qubit-allocation` step).
+      Providing as input the following IR :
 
       ```mlir
-      %0 = "arith.constant"() <{value = dense<3> : tensor<i64>}> : () -> tensor<i64>
-      %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)
+        %0 = "arith.constant"()<{value = dense<3>: tensor<i64>}> : ()->tensor<i64>
+        %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) ->(!jasp.QubitArray, !jasp.QuantumState)
```
```suggestion
      %0 = "arith.constant"()<{value = dense<3>: tensor<i64>}> : () -> tensor<i64>
      %1, %2 = "jasp.create_qubits"(%0, %arg14) : (tensor<i64>, !jasp.QuantumState) ->(!jasp.QubitArray, !jasp.QuantumState)
```

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
```diff
@@ -49,34 +49,65 @@ def JaspToQC : Pass<"jasp-to-qc"> {
...
 
       the pass will translate it into:
 
       ```mlir
-      %0 = qc.static 0 : !qc.qubit
-      %1 = qc.static 1 : !qc.qubit
-      %2 = qc.static 2 : !qc.qubit
+        %0 = qc.static 0 : !qc.qubit
+        %1 = qc.static 1 : !qc.qubit
+        %2 = qc.static 2 : !qc.qubit
```
```suggestion
      %0 = qc.static 0 : !qc.qubit
      %1 = qc.static 1 : !qc.qubit
      %2 = qc.static 2 : !qc.qubit
```

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
rainij: Please do not rely on such subtle things. As far as I can tell the spec does not explicitly say it is not allowed. The wording might simply be sloppy.

In contrast there is some evidence that multiple entry points are desired. See for example qir-runner (https://pypi.org/project/qirrunner/). It is developed in the same organization as the spec and it has an option `-e` to select the entry point. I tested it, you can give it a program with multiple entry points and in that case you need to choose via `-e`. If you omit the option the program says

```
RuntimeError: Multiple matching entry points found.
```

This does not sound like an accident. The developers anticipated multiple entry points and added the `-e` option.

@burgholzer any insights into this topic from your side? I could not find any definitive statement.

rainij: Even independent of the QIR question. I think the current behavior is bad if you want to analyze what is going on. If you simplify the code as suggested then the following statement would be true:

> After the pass has executed successfully you know that your IR contains a `@main` function which is an entry point.

Currently this is not true. You only can ensure that _some_ function is an entry point. This is not an ideal situation for subsequent passes.

One might object that having multiple entry points also complicates things. But this is another question altogether. The current implementation also does not guarantee this. If one really needs this in specific scenarios one can still add an additional check on top (somewhere).

rainij (changes requested):

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
burgholzer: As far as I understood the QIR spec, it distinguishes between the notions of a quantum "application" and a quantum "program".
A quantum program is a single entry point to a function containing code that is compliant with some QIR profile.
An application is the larger classical-quantum program that drives a particular workload. It may contain multiple quantum programs and, hence, multiple entry points.

This is kind of similar to a PennyLane program that may contain multiple `@qnode` decorators as part of one bigger program.

Overall, if we want to support compiling full applications, we need to support multiple entry points.

RiccardoRomanello: The logic that I implemented in the last commit is the following:

_add the entry point attribute to "main" (or whatever the option says) no matter if any other function is already entrypoint. But fail if "main" does not exist._

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
rainij: I know but this is OK IMO. The default is `main` and the option just overrides the behavior a bit. The problem with the original summary is that it assumes that you know what an "entry point" is. We have no crystal clear definition. I would just say, the func with the `qcc.entry_point` attribute is the entry point. This would be confusing then. The summary I suggest says it more directly what it does.

rainij (changes requested):
Just one issue left to resolve. @burgholzer can also already give his review.

mlir/include/qcc/Conversion/JaspToQC/JaspToQC.td
RiccardoRomanello: I agree with you, changed the summary.

rainij (approved):

mlir/lib/Conversion/JaspToQC/AddEntrypointToMain.cpp
```diff
@@ -0,0 +1,50 @@
...
+    auto funcOp = moduleOp.lookupSymbol<func::FuncOp>(entryPointName);
+    if (!funcOp) {
+      moduleOp.emitError("could not find entry-point function '") << entryPointName << "'";
+      return signalPassFailure();
+    }
+
+    // Mark it as entry-point. setAttr is idempotent: if the attribute is
+    // already present with the same value, this is a no-op.
+    OpBuilder builder(funcOp.getContext());
+    funcOp->setAttr(qcc::entryPointAttrName, builder.getUnitAttr());
```
burgholzer: Just a really really nitpicky comment: is there really no other way than to construct a builder here for getting the UnitAttr type? 
Feels a bit unnecessary.

burgholzer (approved):
LGTM 👍🏻
One minor nitpick comment. Feel free to ignore and merge.

mlir/lib/Conversion/JaspToQC/AddEntrypointToMain.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    // Find the function with the configured name. Fail if it does not exist.
+    auto funcOp = moduleOp.lookupSymbol<func::FuncOp>(entryPointName);
+    if (!funcOp) {
+      moduleOp.emitError("could not find entry-point function '") << entryPointName << "'";
+      return signalPassFailure();
+    }
+
+    // Mark it as entry-point. setAttr is idempotent: if the attribute is
+    // already present with the same value, this is a no-op.
```
rainij:
```suggestion
    // NOTE: setAttr is idempotent: if the attribute is
    // already present with the same value, this is a no-op.
```

Just saw it. Please avoid unnecessary comments. It is clear from the code that this marks the entry point. The comment on impotence is OK (not necessary though) as one has to work a bit to see this.

mlir/lib/Conversion/JaspToQC/AddEntrypointToMain.cpp
```diff
@@ -0,0 +1,50 @@
...
+namespace {
+
+struct AddEntrypointToMain final : public impl::AddEntrypointToMainBase<AddEntrypointToMain> {
+  using AddEntrypointToMainBase<AddEntrypointToMain>::AddEntrypointToMainBase;
+
+protected:
+  void runOnOperation() override {
+    ModuleOp moduleOp = getOperation();
+
+    // Find the function with the configured name. Fail if it does not exist.
```
rainij:
```suggestion
```

Evident from the code.

mlir/lib/Conversion/JaspToQC/AddEntrypointToMain.cpp
RiccardoRomanello: Updated with a UnitAttr factory + context. Thanks for the suggestion. I investigated and the builder seems like a really heavy object to build.

burgholzer (approved):

rainij (approved):

Issue #50, opened 2026-05-18
Add pass: insert implicit record ops to entrypoint


rainij:
Create a pass which takes IR similar to this:

```mlir
// NOTE: important is that this has the `entrypoint` attribute (name does not matter).
func.func public @main() -> ... attributes { qcc.entry_point } {
  // ...
  return %a, %b : i64, i1
}
```

And replaces it by something like this (details can be carved out in PR):

```mlir
func.func public @main() -> void attributes { qcc.entry_point } {
  // ...
  aux.record_integer %a : i64 // this op needs to be added
  aux.record_bool %b : i1     // this op already exists
  return
}
```

This follows the recipe of [QIR output recording](https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Adaptive_Profile.md#output-recording).

Note that QIR enables one to assemble complex data structures to be recorded. For example to record something like `(1, 2, [true, false])` (tuple of two ints and an array of bools). This would need "tuple record" and "array record" ops / QIR-funcs in case we want to support it.

- [x] Support for integers and bools (see https://github.com/FullStaQD/compiler/pull/70).
- [x] Support for arrays and tuple (recursively). Moved into issue #106.

# Motivation (Qrisp)

Consider the following qrisp (python) example

```python
from qrisp import *
from qrisp.jasp import make_jaspr
import jax.numpy as jnp

def example():
    qv = QuantumVariable(2)
    h(qv[0])
    cx(qv[0], qv[1])
    a = measure(qv) + 5 # becomes i64 (more precisely: tensor<i64>)
    b = True            # becomes i1 (more precisely: tensor<i1>)

    # There are also things like this, but we leave them out in the first iteration as 
    # they get unpacked within the return statement (c, d, e become three times tensor<i64>
    # (each) which is impossible to distinguish from three integer variables).
    c = [1, 2, 3]
    d = (1, 2, 3)
    e = (1, [2, 3])

    return a, b #, c, d, e

jaspr = make_jaspr(example)()
mlir_stablehlo = jaspr.to_mlir()
print(mlir_stablehlo)
```

This gets translated into this mlir:

```mlir
builtin.module @jasp_module {
  func.func public @main(%arg13: !jasp.QuantumState) -> (tensor<i64>, tensor<i1>, !jasp.QuantumState) {
    // create %10 etc here ...
    func.return %12, %13, %10 : tensor<i64>, tensor<i1>, !jasp.QuantumState
  }
...
}
```

[...]

PR #70, opened 2026-06-02
50 output recording

ctfle:
Relates to #50. For now we only support recording of i64 and i1.


mlir/test/tools/qcc-opt/ToQIR/output-recording.mlir
```diff
@@ -0,0 +1,99 @@
+// RUN: qcc-opt %s -integer-output-recording | FileCheck %s
+
+//===----------------------------------------------------------------------===//
+//
+// Integer output recording pass tests
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: qcc-opt %s -integer-output-recording | FileCheck %s
```
rainij:
```suggestion
```

The second RUN command is probably just a typo, the global docstring is not needed as the RUN command documents this.

mlir/test/tools/qcc-opt/ToQIR/output-recording.mlir
```diff
@@ -0,0 +1,99 @@
...
+
+// -----------------------------------------------------------------------------
+// Test 1: Entry point function returning i64 should be transformed
+// -----------------------------------------------------------------------------
```
rainij:
```suggestion
//===----------------------------------------------------------------------===//
// Entry point function returning i64 should be transformed
//===----------------------------------------------------------------------===//
```

to be consistent with the LLVM "block comment" style. Also, I would not enumerate the tests, gets easily wrong if somebody inserts something in the middle.

mlir/test/tools/qcc-opt/ToQIR/output-recording.mlir
```diff
@@ -0,0 +1,99 @@
...
+
+func.func @test_entry_point_i64() -> i64 attributes { qcc.entry_point } {
+  %q5 = qc.static 5 : !qc.qubit
+  %q7 = qc.static 7 : !qc.qubit
+
+  %c = arith.constant 1 : i64
+  %c1 = arith.addi %c, %c : i64
+  return %c1 : i64
```
rainij: I would simplify the test to the absolute minimum. So the body should just contain an `arith.constant` and return the value. I believe the rest does not add any value to the test.

mlir/test/tools/qcc-opt/output-recording.mlir
rainij, line 1: The test file should be not inside "ToQIR". It is definitely related to QIR, but in part also by Qrisp.

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.td
```diff
@@ -0,0 +1,43 @@
...
+// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM
+// Exceptions.
+// See <repo-root>/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// ===----------------------------------------------------------------------===//
+
+include "mlir/Pass/PassBase.td"
+
+def AuxOutputRecording : Pass<"integer-output-recording"> {
```
rainij: I would rename the pass to `convert-return-to-output-recording`. It does not need to be restricted to integers in general.

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.td
```diff
@@ -0,0 +1,43 @@
...
+    let summary = "Add recording of integer outputs.";
+    let description = [{
+        This pass adds `aux.record_integer` operations after every operation that returns an integer value.
+    }];
```
rainij:
```suggestion
```

duplicate.

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.td
```diff
@@ -0,0 +1,43 @@
...
+    let summary = "Adds recording of integer outputs following the QIR recipe for output recording.";
```
rainij:
```suggestion
    let summary = "Replaces return values of entry point function by record ops.";
```

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.td
```diff
@@ -0,0 +1,43 @@
...
+    let description = [{
```
rainij, line 14: This is already a good docstring. I like the example section. Can you add a section somewhere which tells that this is motivated by Qrisp and QIR (and how).

Then also a very brief TODO comment which mentions that we possibly want to extend this in the future to support "structured" return values like `(1, [false, true])`.

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.td
rainij, line 1: The formatting in this file is slightly off. Please check. Unfortunately we have no tooling here.

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
rainij, line 1: FYI: https://github.com/FullStaQD/compiler/issues/71 . If you like you can work on this in a follow up PR (not here). If yes, please assign yourself so that everybody knows.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
+// ===----------------------------------------------------------------------===//
+//
+// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM
+// Exceptions.
+// See <repo-root>/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// ===----------------------------------------------------------------------===//
+
+#include "qcc/Conversion/Aux_/AuxOutputRecording.h"
```
rainij:
```suggestion
#include "qcc/Conversion/Aux_/AuxOutputRecording.h"  // IWYU pragma: keep
```

See https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+namespace {
+
+struct AuxOutputRecording final : public impl::AuxOutputRecordingBase<AuxOutputRecording> {
+
+  using AuxOutputRecordingBase<AuxOutputRecording>::AuxOutputRecordingBase;
+
+protected:
+  void runOnOperation() override {
+    // Load the aux dialect into the context
```
rainij:
```suggestion
```

We follow the convention to only add comment for non-trivial stuff.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+    auto* context = &getContext();
+    context->loadDialect<::qcc::aux::AuxDialect>();
+
+    auto module = cast<mlir::ModuleOp>(getOperation());
+
+    module.walk([&](func::FuncOp funcOp) {
+      if (!funcOp) {
+        return;
+      }
+      // Only transform functions marked as entry points
```
rainij:
```suggestion
```

code explains itself

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+      if (!funcOp->hasAttr("qcc.entry_point")) {
+        return;
+      }
+
+      // We only care if there are results to record
```
rainij:
```suggestion
```

code explains itself

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+      if (funcOp.getNumResults() == 0) {
+        return;
+      }
+
+      // Build a new function type with void return type
+      // Keep the original argument types, drop results.
```
rainij, line 57:
```suggestion
```

code explains itself

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+      auto oldType = funcOp.getFunctionType();
+      auto newType = FunctionType::get(funcOp->getContext(), oldType.getInputs(), {});
+
+      // Update the function type
```
rainij:
```suggestion
```

code explains itself.

There are more comments like this. Please check the rest below and remove the trivial ones (probably most of them).

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // Load the aux dialect into the context
+    auto* context = &getContext();
+    context->loadDialect<::qcc::aux::AuxDialect>();
+
+    auto module = cast<mlir::ModuleOp>(getOperation());
+
+    module.walk([&](func::FuncOp funcOp) {
+      if (!funcOp) {
+        return;
+      }
```
rainij:
```suggestion
```

not needed. `funcOp` will never be null here.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+      // Keep the original argument types, drop results.
+      auto oldType = funcOp.getFunctionType();
+      auto newType = FunctionType::get(funcOp->getContext(), oldType.getInputs(), {});
+
+      // Update the function type
+      funcOp.setType(newType);
+
+      // Work on the first block
+      auto& body = funcOp.getBody().front();
+      auto* returnOp = body.getTerminator();
```
rainij: I think this is actually not correct _in general_. Any block can have a return statement. There are acutally two ways coming to my mind to handle this:

- check that there is only one block and emit a pass failure (and a diagnostic message) if otherwise.
- handle multiple blocks.

Multiple blocks can only happen when we have unstructured control flow like `cf` or `llvm`. IMO we should go the first route (which is also far easier I suppose). This would also require at least a sentence in the tablegen description file. 

Also add a test file which checks for the diagnostic. See `mlir/test/tools/qcc-opt/jasp-check-static-qubit-allocation-test.mlir` for inspiration (`--verify-diagnostics` is the magic ingredient).

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+      auto oldReturnOperands = retOp.getOperands();
+
+      OpBuilder builder(retOp);
+      // If there are no return operands, nothing to record
+      if (oldReturnOperands.empty()) {
+        retOp.erase();
+        builder.setInsertionPointToEnd(&body);
+        func::ReturnOp::create(builder, retOp.getLoc());
+        return;
+      }
```
rainij: I believe this code is dead. An earlier check makes sure of that.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        if (ty.isInteger(64)) {
+          // aux.record_integer %a : i64
+          ::qcc::aux::RecordIntOp::create(builder, loc, v);
+        } else if (ty.isInteger(1)) {
+          // aux.record_bool %b : i1
+          ::qcc::aux::RecordBoolOp::create(builder, loc, v);
+        } else {
+          // TODO: Handle other types as needed. For now, we only support i64 and i1.
+          return signalPassFailure();
```
rainij: This should be outside the walk. Instead the walk should return a walk result (advance vs interrupt) and based on the result you do 

```cpp
if (result.wasInterrupted())
  return signalPassFailure();
```

_after_ the walk. At least this is idiomatic. One issue with the current implementation is that the pass still runs till the end. The pass itself does not return at this point it only sets a flag that it is failed.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+struct AuxOutputRecording final : public impl::AuxOutputRecordingBase<AuxOutputRecording> {
+
+  using AuxOutputRecordingBase<AuxOutputRecording>::AuxOutputRecordingBase;
+
+protected:
+  void runOnOperation() override {
+    // Load the aux dialect into the context
+    auto* context = &getContext();
+    context->loadDialect<::qcc::aux::AuxDialect>();
```
rainij: Is this really needed (loading AuxDialect)?

rainij (changes requested):

mlir/test/tools/qcc-opt/ToQIR/output-recording.mlir
ctfle: cleaned up 👍

mlir/test/tools/qcc-opt/ToQIR/output-recording.mlir
ctfle: fixed

mlir/test/tools/qcc-opt/ToQIR/output-recording.mlir
ctfle: fixed

mlir/test/tools/qcc-opt/output-recording.mlir
ctfle: 👍

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.td
ctfle: 👍

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
ctfle: removed

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
ctfle: added a check for single block

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
ctfle: right. removed.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
ctfle: Yes thats better. Changed.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
ctfle: No. Seems like tablegen already takes care of that. Removed

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
rainij: The cpp code is fine now, please also document this as an assumption of the pass (in td file).

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
rainij: One small detail. Currently it is

```cpp
    if (walkResult.wasInterrupted()) {
      signalPassFailure();
    }
```

in your code. Better use 

```cpp
    if (walkResult.wasInterrupted()) {
      return signalPassFailure();
    }
```

This is idiomatic LLVM.

rainij (changes requested):

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -0,0 +1,103 @@
...
+      // Record each return value according to its type
+      for (Value v : oldReturnOperands) {
+        Type ty = v.getType();
+
+        if (ty.isInteger(64)) {
+          // aux.record_integer %a : i64
+          ::qcc::aux::RecordIntOp::create(builder, loc, v);
+        } else if (ty.isInteger(1)) {
+          // aux.record_bool %b : i1
+          ::qcc::aux::RecordBoolOp::create(builder, loc, v);
```
RiccardoRomanello: Both the `create` ops can be called just via `aux::`, since the class itself is in the `qcc` scope.

RiccardoRomanello (changes requested):
Just me being picky

RiccardoRomanello (approved):
Everything looks good to me

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
ctfle: ok. I removed most of the comments.

rainij (approved):
LGTM

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.h
burgholzer, line 1: Nitpick, but why the duplication of aux in the directory and the filename? (Holds for all related naming)

burgholzer (approved):
LGTM 👍🏻
Also here just a single nitpick. Feel free to ignore.

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.h
ctfle: From the repo it looked like convention to add the dialect in related filenames. No big deal to change that (but I guess these files are temporary anyways).

rainij: @ctfle I am undecided on this. If you are too then just copy the behavior of LLVM. Then ask for this to be merged either by me or Lukas.

ctfle: For me this is fine as is.

PR #73, opened 2026-06-08
Adding RZ QIR to Quantum Instruction Set (QIS)

### LINKED ISSUE #41: Add RZ to QIR QIS
The corresponding QIS function will probably look like this

```llvm
declare void @__quantum__qis__rz__body(double, ptr)
```

# Original suggestion (DO NOT IMPLEMENT!)

Originally we wanted an fixed point angle with this format:

```llvm
declare void @__quantum__qis__rz__body(i32, ptr)
```

The `i32` encodes an angle from the interval $[0,2\pi)$. Although the hardware seems to prefer fixed point value for efficiency the QIR community seems to have settled on floats (see discussion below). This probably introduces some friction, especially with high speed modalities but we handle this in upcoming issues.

RiccardoRomanello:
Resolves #41


mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -188,11 +189,22 @@ struct UnitaryLowering : public ConversionPattern {
...
-    auto allPtrs = qubitsToPtrs(rewriter, unitaryOp.getControls());
+    // Per QIR convention, non-qubit parameters (e.g. rotation angles) come first,
+    // followed by qubit pointers. For non-parametric gates (X, H, CX) there are no
+    // non-qubit operands so behaviour is identical to before.
+    SmallVector<Value> args;
+    for (Value operand : op->getOperands()) {
+      if (!isa<qc::QubitType>(operand.getType())) {
+        args.push_back(operand);
+      }
+    }
```
rainij: Is there some interface which would directly get you the "parameters"? Maybe @burgholzer knows more. If not then leave as is.

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
```diff
@@ -154,6 +155,23 @@ struct PrepToQIR final : public impl::PrepToQIRBase<PrepToQIR> {
...
+  /// rotation angle (double-precision float) and second is the target qubit pointer.
+  void createF64PtrFnDecl(StringRef fnName) {
+    ModuleOp moduleOp = getOperation();
+    auto* ctx = moduleOp.getContext();
+    OpBuilder builder(ctx);
+    builder.setInsertionPointToEnd(moduleOp.getBody());
+
+    auto f64Type = builder.getF64Type();
+    auto ptrType = LLVM::LLVMPointerType::get(ctx);
+    auto fnType = LLVM::LLVMFunctionType::get(LLVM::LLVMVoidType::get(ctx), {f64Type, ptrType});
```
rainij, line 192: Please always check the CI before submitting review. It flags the `f64Type`. You should also see it in your IDE if it is set up correctly.

```
Slicing object from type 'FloatType' to 'Type' discards 8 bytes of state
```

rainij (changes requested):
Looks almost good to go. Just a few tiny things.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
burgholzer: The UnitaryOpInterface exposes dedicated methods for this; see https://github.com/munich-quantum-toolkit/core/blob/611e27c3af52cb016bb5c85fc2ad6d15bbdff821/mlir/include/mlir/Dialect/QC/IR/QCInterfaces.td#L57-L58 and the rest of the file.

RiccardoRomanello: I used unitaryOp to obtain the parameters! Thanks for the suggestion.

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
RiccardoRomanello: Thanks, the clang-tidy issue made me realize linting was not working anymore in my VSCode. 
Fixed and updated the code to avoid the code. Had to go through opaquePointers

rainij (approved):
LGTM. Can you investigate the windows fail in the CI?

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -188,11 +189,14 @@ struct UnitaryLowering : public ConversionPattern {
       return emitMissingQIRDeclError(unitaryOp, qisName);
     }
 
-    auto allPtrs = qubitsToPtrs(rewriter, unitaryOp.getControls());
+    SmallVector<Value> args(unitaryOp.getParameters().begin(), unitaryOp.getParameters().end());
```
burgholzer:
```suggestion
    auto args = llvm::to_vector(unitaryOp.getParameters());
```
Is likely more idiomatic here.

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
```diff
@@ -154,6 +155,25 @@ struct PrepToQIR final : public impl::PrepToQIRBase<PrepToQIR> {
...
+    auto* ctx = moduleOp.getContext();
+    OpBuilder builder(ctx);
+    builder.setInsertionPointToEnd(moduleOp.getBody());
+
+    auto f64Type = builder.getF64Type();
+    auto ptrType = LLVM::LLVMPointerType::get(ctx);
+
+    auto fnType = LLVM::LLVMFunctionType::get(LLVM::LLVMVoidType::get(ctx),
+                                              {mlir::Type::getFromOpaquePointer(f64Type.getAsOpaquePointer()),
+                                               mlir::Type::getFromOpaquePointer(ptrType.getAsOpaquePointer())});
```
burgholzer: This doesn't quite feel right. I suppose this was a fix for the slicing clang-tidy warning.
That warning seems to be a false positive as the F64Type is used in other places across the LLVM codebase just like that.
I'd rather argue for locally disabling the clang-tidy check with a `// NOLINT(...)` comment.

burgholzer (changes requested):
Two minor requests for changes to clean up the code.

The Windows failure is likely due to GitHub's "interesting" move to update the default MSVC toolset being used on the `windows2025` runners.
Apparently, the new MSVC version has a problem with one of the calls in a particular file. Should be fairly straight forward to fix based on the error log from the CI.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
RiccardoRomanello: Applied the suggestion

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
RiccardoRomanello: I applied the NOLINT comment

RiccardoRomanello:
For what concerns the windows issue, the problem is windows not locating a `.lib` file correlated to _DIA SDK_. 
Since I am really not used to these kind of settings, I asked to Claude on GitHub for some help. He told me to disallow the dependency in CMake if we do not do it. Otherwise, another approach is to change the `ci.yml` file and have the cpp-tests-windows job to install the DIA SDK before running the tests. Which path should we take?

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
burgholzer: Please always use specific rules for the linter exclusion comments.
These blanked ignores ignore _all_ rules for the particular line, which is not what is intended here.

burgholzer:
@denialhaag We saw this Windows issue with DIA in MQT Core as well for a brief time; I can't quite remember what we did to work around that. Do you have an idea about that?
This definitely has to do with the MSVC change. It is looking in the wrong directory for the DIA library.
Maybe worth trying to bump the LLVM version in the CI? Latest would be 21.1.7

RiccardoRomanello:
@burgholzer From the CI file, this is the current version of LLVM adopted: llvm-version: 22.1.0

Are you suggesting to lower it?

Moreover, is there a way to test those settings locally without having to push and 'fingercrossing' that everything works?

Thanks

denialhaag:
> @burgholzer From the CI file, this is the current version of LLVM adopted: llvm-version: 22.1.0
> 
> Are you suggesting to lower it?
> 
> Moreover, is there a way to test those settings locally without having to push and 'fingercrossing' that everything works?
> 
> Thanks

Sorry for not responding here. I'm doing the update in #78.

rainij:
@burgholzer can this be merged?

burgholzer:
> @burgholzer can this be merged?

Conflicts need to be resolved and there is still an open conversation left to be addressed.

mlir/lib/Conversion/ToQIR/PrepToQIR.cpp
RiccardoRomanello: Sorry I completely missed the comment. I updated the file.

RiccardoRomanello:
Sorry, I completely missed the comment on `NOLINT` in the middle of the microsoft testing thingy. I updated with the suggestions.

burgholzer:
> Sorry, I completely missed the comment on `NOLINT` in the middle of the microsoft testing thingy. I updated with the suggestions. 

No worries.
You still need to resolve the conflicts before this can be merged.

rainij:
@RiccardoRomanello what is the status of this PR?

RiccardoRomanello:
@rainij Just resolved the conflicts.

burgholzer (approved):

PR #74, opened 2026-06-08
71 unify integer output recording

### LINKED ISSUE #71: Unify integer output recording
We should unify `record_bool` and `record_integer` to a single operation. A bool is also just an integer. The new op should accept _any integer_ as operand:

```mlir
aux.record_int %c1 : i1
aux.record_int %c1 : i64
```

Further down the QIR exporter can select the right runtime function based on the datatype.

ctfle:
Resolves #71


rainij (changes requested):
This PR contains the changes of https://github.com/FullStaQD/compiler/pull/70/changes#diff-ddf324d534ea8a1cc76b6e77c5e2336a4cf46ff24018367f8ad5a2cd50117c70 . Hence a review does not make sense right now. Please wait until the other one is merged if this cannot be done independently. But you could fix the linter errors in the mean-time.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -151,17 +149,27 @@ struct MeasureLowering : public OpConversionPattern<qc::MeasureOp> {
...
-    LLVM::CallOp::create(rewriter, loc, TypeRange(), qirRtBoolRecordOutput, ValueRange{adaptor.getValue(), addressOf});
+
+    auto operand = op->getOperand(0);
+    Type ty = operand.getType();
+
+    llvm::StringRef callee;
+    if (ty.isInteger(1)) {
+      callee = qirRtBoolRecordOutput;
+    } else {
+      callee = qirRtIntRecordOutput;
```
RiccardoRomanello, line 169: Maybe it is just me being picky, but I would rather check first if `isInteger(1)` is false, and so apply the `qirRtIntRecordOutput` in the true branch of the main. Leave the `qirRtBoolRecordOutput` for the false branch. 

This is mostly because the pass is for int recording, so I would give the 'most important path' (true one) to int and the least important one (false) to bool

RiccardoRomanello (changes requested):
Looks almost good to me, just one minor change.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
ctfle: Ok I see the point. From the clean code perspective its probably better to keep the conditions as simple as possible (not adding negations). But it really doesn't matter that much. I changed it 👍

RiccardoRomanello (approved):
Looks good to me now

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -151,17 +149,27 @@ struct MeasureLowering : public OpConversionPattern<qc::MeasureOp> {
...
+
+    auto operand = op->getOperand(0);
+    Type ty = operand.getType();
+
+    llvm::StringRef callee;
+    if (!ty.isInteger(1)) {
+      callee = qirRtIntRecordOutput;
+    } else {
+      callee = qirRtBoolRecordOutput;
+    }
```
rainij, line 172: According to the `Constants.h` file the signature for the integer recording is `void(i64,ptr)`. So the correct move would be to match only if the integer width is 1 or 64 (return failure otherwise).

If we had an integer with width of, say, 32 the pass should then fail. It is already a conversion pass and declares e.g. QC as illegal. It probably does not declare Aux dialect as illegal, right now. But IMO we should. Should be basically one line to add.

mlir/test/tools/qcc/example_ghz.mlir
```diff
@@ -56,8 +56,66 @@ func.func @ghz_manual() attributes { qcc.entry_point } {
 // CHECK:           llvm.return
 // CHECK:         }
 
+func.func @ghz_manual_with_int_record_output() attributes { qcc.entry_point } {
```
rainij: Please delete one of the two tests. IMO we can keep your extended test.

You might also want to look into the `--variable_names` of the generator script for the CHECK lines. This way we could provide some easier to read names. E.g. for the constants names which indicate the value.

rainij (changes requested):
Looks almost good to go. Just a few small things left to do.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
ctfle: Ok. ~Another option would be to add integers of different width to the recording (also one line).~ Never mind, `__quantum__rt__int_record_output` wants to have an i64

ctfle: ok I added an explicit check for i64 and marked Aux as illegal.

mlir/test/tools/qcc/example_ghz.mlir
ctfle: 👍

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir.mlir
```diff
@@ -63,11 +63,12 @@ func.func @test() -> i64 attributes { qcc.entry_point } {
 // The pass assumes that these decls already exist.
 llvm.func @__quantum__rt__initialize(!llvm.ptr)
 llvm.func @__quantum__rt__bool_record_output(i1, !llvm.ptr)
+llvm.func @__quantum__rt__integer_record_output(i64, !llvm.ptr)
```
rainij:
```suggestion
llvm.func @__quantum__rt__int_record_output(i64, !llvm.ptr)
```

This also indicates that there is no test which uses this QIR function. Please add two lines to the file exercising this path.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -69,11 +69,8 @@ struct AuxOutputRecording final : public impl::AuxOutputRecordingBase<AuxOutputR
...
       for (Value v : oldReturnOperands) {
         Type ty = v.getType();
-
-        if (ty.isInteger(64)) {
+        if (ty.isInteger()) {
           aux::RecordIntOp::create(builder, loc, v);
-        } else if (ty.isInteger(1)) {
-          aux::RecordBoolOp::create(builder, loc, v);
         } else {
           // TODO: Handle other types as needed. For now, we only support i64 and i1.
```
rainij:
```suggestion
          // TODO: Add support for other types as needed.
```

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -151,17 +149,30 @@ struct MeasureLowering : public OpConversionPattern<qc::MeasureOp> {
...
                                 ConversionPatternRewriter& rewriter) const override {
     auto loc = op.getLoc();
     StringRef labelName = qcc::qirDummyLabelGlobalSymbolName;
 
     auto addressOf =
         LLVM::AddressOfOp::create(rewriter, loc, LLVM::LLVMPointerType::get(rewriter.getContext()), labelName);
-    LLVM::CallOp::create(rewriter, loc, TypeRange(), qirRtBoolRecordOutput, ValueRange{adaptor.getValue(), addressOf});
+
+    auto operand = op->getOperand(0);
+    Type ty = operand.getType();
```
rainij:
```suggestion
    Type ty = op.getValue().getType();
```

mlir/test/tools/qcc-opt/output-recording.mlir
```diff
@@ -81,7 +81,7 @@ func.func @test_multiple_returns() -> (i64, i1) attributes { qcc.entry_point } {
 // CHECK:     %[[constant0:.*]] = arith.constant 2 : i64
 // CHECK:     %[[constant1:.*]] = arith.constant true
 // CHECK:     aux.record_integer %[[constant0]]
```
rainij:
```suggestion
// CHECK:     aux.record_integer %[[constant0]] : i64
```

mlir/test/tools/qcc-opt/output-recording.mlir
```diff
@@ -81,7 +81,7 @@ func.func @test_multiple_returns() -> (i64, i1) attributes { qcc.entry_point } {
...
-// CHECK:     aux.record_bool %[[constant1]]
+// CHECK:     aux.record_integer %[[constant1]]
```
rainij:
```suggestion
// CHECK:     aux.record_integer %[[constant1]] : i1
```

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir.mlir
rainij: In the light of this I would advocate for renaming the operation to `aux.record_int`. I already wanted to comment on this when it was added. Now it feels a bit justified. If you are against it feel free to reject the idea.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -151,17 +149,30 @@ struct MeasureLowering : public OpConversionPattern<qc::MeasureOp> {
...
+
+    llvm::StringRef callee;
+    if (ty.isInteger(1)) {
+      callee = qirRtBoolRecordOutput;
+    } else if (ty.isInteger(64)) {
+      callee = qirRtIntRecordOutput;
+    } else {
+      return op.emitError("unsupported type for recording output: ") << ty;
```
rainij:
```suggestion
      return failure();
```

I think this is cleaner. The pass itself already treats aux dialect as illegal. This here would just lead to a double diagnostic.

rainij (changes requested):
I found a few more (tiny) things to fix (one is about test coverage). There are also merge conflicts now.

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir.mlir
ctfle: Yes makes sense 👍

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.td
```diff
@@ -10,9 +10,9 @@
 include "mlir/Pass/PassBase.td"
 
 def AuxOutputRecording : Pass<"convert-return-to-output-recording"> {
-    let summary = "Replaces return values of entry point function by record ops.";
-    let description = [{
-      This pass replaces integer returns of the entry point function with `aux.record_integer` and `aux.record_bool` and
+  let summary = "Replaces return values of entry point function by record ops.";
+  let description = [{
+      This pass replaces integer returns of the entry point function with `aux.record_int` and `aux.record_bool` and
```
rainij: Contains still a reference to `aux.record_bool`.

mlir/test/tools/qcc-opt/output-recording.mlir
```diff
@@ -16,7 +16,7 @@ func.func @test_entry_point_i64() -> i64 attributes { qcc.entry_point } {
 }
 // CHECK-LABEL:   func.func @test_entry_point_i64() attributes {qcc.entry_point} {
 // CHECK:     %[[constant:.*]] = arith.constant 1 : i64
-// CHECK:     aux.record_integer %[[constant]]
+// CHECK:     aux.record_int %[[constant]]
```
rainij:
```suggestion
// CHECK:     aux.record_int %[[constant]] : i64
```

mlir/test/tools/qcc-opt/output-recording.mlir
```diff
@@ -46,8 +46,8 @@ func.func @test_multiple_i64() -> (i64, i64) attributes { qcc.entry_point } {
 // CHECK-LABEL:   func.func @test_multiple_i64() attributes {qcc.entry_point} {
 // CHECK:     %[[constant0:.*]] = arith.constant 1 : i64
 // CHECK:     %[[constant1:.*]] = arith.constant 2 : i64
-// CHECK:     aux.record_integer %[[constant0]]
-// CHECK:     aux.record_integer %[[constant1]]
+// CHECK:     aux.record_int %[[constant0]]
+// CHECK:     aux.record_int %[[constant1]]
```
rainij:
```suggestion
// CHECK:     aux.record_int %[[constant0]] : i64
// CHECK:     aux.record_int %[[constant1]] : i64
```

rainij (changes requested):
Just 3 tiny things, should be a matter of minutes to fix. Then good to go.

burgholzer (approved):
LGTM

rainij (approved):

PR #78, opened 2026-06-11
Update workflows to use LLVM 22.1.7

denialhaag:
This PR updates [munich-quantum-toolkit/workflows](https://github.com/munich-quantum-toolkit/workflows) to the latest version and updates the workflows to use LLVM 22.1.7.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ permissions:
 jobs:
   change-detection:
     name: 🔍 Change
-    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-change-detection.yml@143bf51e7192c31e76abafe58e3732c5aff9c69b # v2.0.1
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-change-detection.yml@eeea704310e1f9753ce49ac9a276145121f4713c # v2.1.0
 
   cpp-tests-ubuntu:
     name: 🇨‌ Test 🐧
@@ -33,12 +33,12 @@ jobs:
           - runs-on: ubuntu-24.04
             compiler: gcc
             preset: debug
-    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-tests-ubuntu.yml@143bf51e7192c31e76abafe58e3732c5aff9c69b # v2.0.1
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-tests-ubuntu.yml@eeea704310e1f9753ce49ac9a276145121f4713c # v2.1.0
     with:
       runs-on: ${{ matrix.runs-on }}
       compiler: ${{ matrix.compiler }}
       setup-mlir: true
-      llvm-version: 22.1.0
+      llvm-version: 22.1.7
       preset-name: ${{ matrix.preset }}
 
   cpp-tests-macos:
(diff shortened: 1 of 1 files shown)


burgholzer (approved):

burgholzer:
We did not switch to 2022 in MQT Core, right?
Maybe we changed something in the CMakePresets (Generator for example)?

denialhaag:
> We did not switch to 2022 in MQT Core, right?

No, we are still using `windows-2025` in MQT Core. I just wanted to test if switching to `windows-2022` does the trick.

> Maybe we changed something in the CMakePresets (Generator for example)?

I just checked again, but the presets are virtually identical. The only difference is that we don't configure with `LLVM_EXTERNAL_LIT` in MQT Core.

---

It looks like the error occurs while linking `qcc`. I don't immediately see the issue, but maybe something is not configured correctly there. I honestly don't remember seeing this error in MQT Core. Did you tweak which libraries we are linking against at some point? 🤔

burgholzer:
Hm. Maybe comparing the qcc and mqt-cc CmakeLists.txt reveals something.

I definitely saw this before.. would have to check the failed job logs in MQT Core over the last couple of weeks.

https://stackoverflow.com/questions/44481648/llvm-cmake-install-cannot-find-dia-sdk provides a local fix. But that cannot be it.

https://stackoverflow.com/questions/78085548/building-with-llvm-using-cmake-llvm-clang-vs-path-is-wrong reports similar problems.

There's also https://github.com/llvm/llvm-project/issues/86250 but no solution yet.

denialhaag:
@burgholzer, looks like everything works if we use `windows-2025` (which is now aliasing `windows-2025-vs2026`) and LLVM 22.1.5, as the latter was also built with the `windows-2025-vs2026` image. 🙃

.github/workflows/ci.yml
```diff
@@ -91,14 +91,14 @@ jobs:
...
-    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@143bf51e7192c31e76abafe58e3732c5aff9c69b # v2.0.1
+    uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@eeea704310e1f9753ce49ac9a276145121f4713c # v2.1.0
     with:
       clang-version: 22
       build-project: true
       files-changed-only: true
       cpp-linter-extra-args: "-std=c++20"
       setup-mlir: true
-      llvm-version: 22.1.0
+      llvm-version: 22.1.5
```
rainij: Why not `22.1.7`? I would go with the newest release if there is no obstacle.

rainij (changes requested):
Would be nice to update. Did you also test that local setup works OK? I ask because I am unsure what exactly the CI tests. If you use devcontainer would be good to test if that one works.

.github/workflows/ci.yml
burgholzer: We are still working some stuff out related to the latest Windows GitHub runner changes. This will likely be bumped to the latest patch release once we got things worked out. We'll keep this open for now.
A previous run with 22.1.7 (https://github.com/FullStaQD/compiler/pull/78/commits/d274e671776f582fdd2054e5a865506f8b56fc32) was already successful on non-Windows platforms.

denialhaag: This is fixed now! 🙂

burgholzer (approved):
Nice 😊

rainij (approved):

PR #79, opened 2026-06-12
Discussion: CMake Problem

julfarn:
My build fails unless I add these specific lines. This is for CMake targets `all` and `test-qcc-project` alike.

### DIFF
--- a/mlir/lib/Compiler/CMakeLists.txt
+++ b/mlir/lib/Compiler/CMakeLists.txt
@@ -1,7 +1,8 @@
 add_mlir_library(
   QCCPipeline
   Pipeline.cpp
-  DEPENDS # FIXME: What's the proper place to put these dependencies? Why do only I need them?
+  DEPENDS
+  AuxOutputRecordingIncGen
   JaspToQCIncGen
   QCCToQIRIncGen
   LINK_LIBS
--- a/mlir/lib/Compiler/CMakeLists.txt
+++ b/mlir/lib/Compiler/CMakeLists.txt
@@ -1,6 +1,9 @@
 add_mlir_library(
   QCCPipeline
   Pipeline.cpp
+  DEPENDS # FIXME: What's the proper place to put these dependencies? Why do only I need them?
+  JaspToQCIncGen
+  QCCToQIRIncGen
   LINK_LIBS
   MLIRPass
   MLIRTransforms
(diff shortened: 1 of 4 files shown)


burgholzer:
Which system are you on? (OS, compiler, etc.).
The proper place for these would probably be the individual targets that need these (`QCCJaspToQC` and `QCCToQIR`)

rainij:
@julfarn confirmed that this is the right fix. I also added some other missing dependencies. Please check that everything is fine on your side. Please also go over the changes I made briefly to double check. If fine then just merge without further approval.

Background info: `DEPENDS` is anything the target needs to be _compiled_. In contrast _LINK_LIBS_ is anything it needs to be linked (later). An IncGen target (generated by tablegen) is needed in DEPENDS if the relevant source (transitively) include the generated header.

This is documented at `function(llvm_add_library name)` which gets these args forwarded from `add_mlir_library`:

```
#   DEPENDS targets...
#     Same semantics as add_dependencies().

#   LINK_LIBS lib_targets...
#     Same semantics as target_link_libraries().
```

See also: https://cmake.org/cmake/help/latest/ .

rainij (approved):

rainij:
> Which system are you on? (OS, compiler, etc.). The proper place for these would probably be the individual targets that need these (`QCCJaspToQC` and `QCCToQIR`)

I did not see your comment. @julfarn maybe then take some care with checking if everything is alright and ask @burgholzer for a final review. If something was not alright with my commit also ping me so that I learn about it. I leave my approval as is.

burgholzer:
> > Which system are you on? (OS, compiler, etc.). The proper place for these would probably be the individual targets that need these (`QCCJaspToQC` and `QCCToQIR`)
> 
> I did not see your comment. @julfarn maybe then take some care with checking if everything is alright and ask @burgholzer for a final review. If something was not alright with my commit also ping me so that I learn about it. I leave my approval as is.

No worries. Your changes look just fine.
I was briefly thinking whether some of the depends statements might be redundant; but at the same time better safe than sorry.
Feel free to simply go ahead 😌

julfarn:
I don't feel comfortable merging this at the moment. It feels wrong to have the inc's explicitly listed as dependencies so far up in the tree. 

@burgholzer I'm on ubuntu 24.4 using gcc-13 (our dev container), and noone else seems to have this problem.

In addition, these are not new features and it used to run fine. I am really puzzled by what could cause this. If it works in a fresh build for at least someone, I don't see why new dependecies should be necessary.

rainij:
> In addition, these are not new features and it used to run fine. I am really puzzled by what could cause this. If it works in a fresh build for at least someone, I don't see why new dependecies should be necessary.

@julfarn I believe those errors are non-deterministic. On my machine the build order might just be different. With DEPENDS we impose constraints on the build order. My order and everybody elses order was probably just OK by accident. But if you still feel not good about it, you can share your error message here.

julfarn:
Ok, I'm also not in the mood to dig that deep into it if it's really non-deterministic. 

I suppose it doesn't hurt to have some redundancy here.

Issue #75, opened 2026-06-08
Qrisp teleport circuit to QIR


rainij:
Make sure the teleport circuit can be compiled to qrisp. Possible obstacles:

- Missing gates in QIR exporter. Add them as needed.
- `scf.if` not lowered. Lower it then.

Beware of https://github.com/FullStaQD/compiler/issues/50 (PR will likely soon be merged).

- [x] Implement the functionality (by #80).
- [x] Implement tests once #81 is merged (c.f. https://github.com/FullStaQD/compiler/pull/80#discussion_r3413436709). (By #104)

# Shell script

The following script should work:

```shell 
# currently with custom qrisp dependency:
python teleport.0.py > teleport.1.mlir

# run our pipeline, lower to llvm/QIR
qcc teleport.1.mlir -o teleport.2.mlir
mlir-translate teleport.2.mlir -mlir-to-llvmir -o teleport.3.ll

# Simulate via qir-runner:
qir-runner -f teleport.3.ll -s 2
```

# Qrisp program (`teleport.0.py`)

```python
from qrisp import QuantumVariable, h, cx, q_cond, s, t, t_dg, measure, control
from qrisp.jasp import make_jaspr

def t_gate_via_magic_state():
    qv = QuantumVariable(2)

    h(qv[0])
    t(qv[0])  # should be magic state

    cx(qv[0], qv[1])
    
    m0 = measure(qv[0])
    q_cond(m0 == 1, lambda q: (s(q), None)[1], lambda q: None, qv[1])

    # Just to check for correctness:
    t_dg(qv[1])
    m1 = measure(qv[1])
    
    return m1 # should be zero always

jaspr = make_jaspr(t_gate_via_magic_state)()
print(jaspr.to_mlir(lower_stablehlo=True))
```

EDIT: I just realized the program is a bit silly. The T gate would not do anything to the zero state in the data qubit anyway. Feel free to modify the script slightly to turn the last measurement into a meaningful check if the rest works.

PR #80, opened 2026-06-12
75 qrisp to qir

ctfle:
Relates to #75 (proper tests follow in a separate PR). Adds the missing gates and lowers scf to llvm.

### DIFF
--- a/mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
+++ b/mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
@@ -262,7 +262,9 @@ struct ConvertJaspQuantumGateOp final : OpConversionPattern<jasp::QuantumGateOp>
     TRY_CONVERT_GATE("rz", qc::RZOp, 0, 1, 1);
     TRY_CONVERT_GATE("crz", qc::RZOp, 1, 1, 1);
     TRY_CONVERT_GATE("s", qc::SOp, 0, 1, 0);
+    TRY_CONVERT_GATE("s_dg", qc::SdgOp, 0, 1, 0);
     TRY_CONVERT_GATE("t", qc::TOp, 0, 1, 0);
+    TRY_CONVERT_GATE("t_dg", qc::TdgOp, 0, 1, 0);
     TRY_CONVERT_GATE("sx", qc::SXOp, 0, 1, 0);
     TRY_CONVERT_GATE("swap", qc::SWAPOp, 0, 2, 0);
     TRY_CONVERT_GATE("rxx", qc::RXXOp, 0, 2, 1);
--- a/mlir/lib/Conversion/ToQIR/FinalizeToQIR.cpp
+++ b/mlir/lib/Conversion/ToQIR/FinalizeToQIR.cpp
@@ -20,6 +20,9 @@
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
+#include <mlir/Conversion/ControlFlowToLLVM/ControlFlowToLLVM.h>
+#include <mlir/Conversion/SCFToControlFlow/SCFToControlFlow.h>
+
 using namespace mlir;
 
 namespace {
@@ -74,6 +77,9 @@ protected:
       RewritePatternSet patterns(ctx);
 
       populateFuncToLLVMConversionPatterns(typeConverter, patterns);
+
+      populateSCFToControlFlowConversionPatterns(patterns);
+      cf::populateControlFlowToLLVMConversionPatterns(typeConverter, patterns);
 
       if (failed(applyFullConversion(moduleOp, target, std::move(patterns)))) {
         return signalPassFailure();
--- a/mlir/test/tools/qcc/example_teleport.mlir
+++ b/mlir/test/tools/qcc/example_teleport.mlir
@@ -35,7 +35,6 @@ builtin.module @jasp_module {
 // CHECK:    %[[CONSTANT_0:.*]] = llvm.mlir.constant(0 : i64) : i64
 // CHECK:    %[[ZERO:.*]] = llvm.mlir.zero : !llvm.ptr
 // CHECK:    llvm.call @__quantum__rt__initialize(%[[ZERO]]) : (!llvm.ptr) -> ()
-
 // CHECK:    %[[INTTOPTR_0:.*]] = llvm.inttoptr %[[CONSTANT_0]] : i64 to !llvm.ptr
 // CHECK:    llvm.call @__quantum__qis__h__body(%[[INTTOPTR_0]]) : (!llvm.ptr) -> ()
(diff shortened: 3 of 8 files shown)


mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -262,7 +262,9 @@ struct ConvertJaspQuantumGateOp final : OpConversionPattern<jasp::QuantumGateOp>
     TRY_CONVERT_GATE("rz", qc::RZOp, 0, 1, 1);
     TRY_CONVERT_GATE("crz", qc::RZOp, 1, 1, 1);
     TRY_CONVERT_GATE("s", qc::SOp, 0, 1, 0);
+    TRY_CONVERT_GATE("s_dg", qc::SdgOp, 0, 1, 0);
     TRY_CONVERT_GATE("t", qc::TOp, 0, 1, 0);
+    TRY_CONVERT_GATE("t_dg", qc::TdgOp, 0, 1, 0);
```
julfarn, line 267: Alternatively, there's an inversion op in `qc`. But I suspect that usind the direct ops when available is preferred to `qc.inv { qc.t ... }`.

mlir/test/tools/qcc/example_teleport.mlir
```diff
@@ -0,0 +1,68 @@
+// RUN: qcc -o - %s | FileCheck %s
+
+// A simple teleportation program, written in jasp and compiled to mlir.
```
julfarn:
```suggestion
// A simple teleportation program, written in qrisp and compiled to mlir.
```

mlir/test/tools/qcc/example_teleport.mlir
```diff
@@ -0,0 +1,68 @@
...
+// CHECK:    llvm.call @__quantum__qis__t__body(%[[INTTOPTR_0]]) : (!llvm.ptr) -> ()
+// CHECK:    %[[INTTOPTR_1:.*]] = llvm.inttoptr %[[CONSTANT_1]] : i64 to !llvm.ptr
+// CHECK:    llvm.call @__quantum__qis__cx__body(%[[INTTOPTR_0]], %[[INTTOPTR_1]]) : (!llvm.ptr, !llvm.ptr) -> ()
+// CHECK:    llvm.call @__quantum__qis__mz__body(%[[INTTOPTR_0]], %[[INTTOPTR_0]]) : (!llvm.ptr, !llvm.ptr) -> ()
+// CHECK:    %[[READ_RESULT_0:.*]] = llvm.call @__quantum__rt__read_result(%[[INTTOPTR_0]]) : (!llvm.ptr) -> i1
+// CHECK:    llvm.cond_br %[[READ_RESULT_0]], ^bb1, ^bb2
+// CHECK:  ^bb1:  // pred: ^bb0
+// CHECK:    llvm.call @__quantum__qis__s__body(%[[INTTOPTR_1]]) : (!llvm.ptr) -> ()
+// CHECK:    llvm.br ^bb2
+// CHECK:  ^bb2:  // 2 preds: ^bb0, ^bb1
```
julfarn, line 50: Are the comments here output by mlir?

mlir/lib/Conversion/ToQIR/FinalizeToQIR.cpp
```diff
@@ -75,6 +78,9 @@ struct FinalizeToQIR final : public impl::FinalizeToQIRBase<FinalizeToQIR> {
 
       populateFuncToLLVMConversionPatterns(typeConverter, patterns);
 
+      populateSCFToControlFlowConversionPatterns(patterns);
+      cf::populateControlFlowToLLVMConversionPatterns(typeConverter, patterns);
```
julfarn, line 82: There's also a dedicated pass `convert-scf-to-cf` which could be inserted into the pipeline instead. 

This feels a bit more natural to me, but it's not a strong opinion at this point. 
If we keep the current approach, please update the docstring in the pass tablegen to mention that scf-to-cf conversion takes place here.

julfarn (changes requested):
Looks good. Good to know that the scf-to-cf conversion was still missing. 

- I would prefer to use the dedicated pass for the conversion.
- Please rebase/merge main so that the pipeline can run. 
- Once (if) #81 gets merged, the same qrisp input could become an end-to-end integration test.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij: Probably a question for @burgholzer .

mlir/lib/Conversion/ToQIR/FinalizeToQIR.cpp
rainij: Would also apply to `populateFuncToLLVMConversionPatterns`. IMO it can stay as is, but we can keep it in mind. Could be that the pass becomes too powerful and we might want to split it up.

mlir/test/tools/qcc/example_teleport.mlir
rainij, line 1: The test is is not so great in the sense that we do not know where it comes from. On the other hand with https://github.com/FullStaQD/compiler/pull/81 there is work under way which makes this possible. Hence my suggestion is as follows:

We split the issue in two parts:
- Implement the functionality with a quick and dirty test (this one).
- Add the proper test once #81 is through.

Does this sound like a plan? If yes I will add two checkboxes to #75 .

rainij (changes requested):
Looks essentially good to me, just a few things to clarify.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
burgholzer: Yeah. `inv { tdg }` will actually be canonicalized to `t` (similar for other ops)

ctfle: Usually especially for T and S its common to have the dagger

mlir/test/tools/qcc/example_teleport.mlir
ctfle: Sounds good 👍

mlir/test/tools/qcc/example_teleport.mlir
ctfle: Yes

mlir/test/tools/qcc/example_teleport.mlir
rainij: Done. For that reason I will change the "resolve" to a "relates" in the PR description.

rainij: Please add a TODO comment referring to the original issue #75  and that we want to replace the test by a proper one.

julfarn (approved):

rainij (approved):

burgholzer (approved):
LGTM 👍🏼

PR #81, opened 2026-06-12
Qrisp integration test framework, ghz tests

### LINKED ISSUE #76: Qrisp GHZ circuit to QIR
Make sure the ghz preparation circuit can be compiled to qrisp. This likely requires converting `scf.while` to `scf.for`, raising to affine dialect and then using the unrolling pass from there.

Beware of https://github.com/FullStaQD/compiler/issues/50 (PR will likely soon be merged).

# Shell script

The following script should work:

```shell 
# currently with custom qrisp dependency:
python ghz.0.py > ghz.1.mlir

# run our pipeline, lower to llvm/QIR
qcc ghz.1.mlir -o ghz.2.mlir
mlir-translate ghz.2.mlir -mlir-to-llvmir -o ghz.3.ll

# Simulate via qir-runner:
qir-runner -f ghz.3.ll -s 2
```

# Qrisp program (`ghz.0.py`)

```python
from qrisp import QuantumVariable, h, cx, measure, q_fori_loop

def prepare_ghz():
    n = 3
    qv = QuantumVariable(n)

    h(qv[0])

    # For now OK to just unroll this loop:
    q_fori_loop(1, n, lambda i, qv: (cx(qv[i-1], qv[i]), qv)[1], qv)

    m = measure(qv)
    
    return m

jaspr = make_jaspr(prepare_ghz)()
print(jaspr.to_mlir(lower_stablehlo=True))
```

NOTE: check that `q_fori_loop` gets turned into a `scf.while`. If we would use a standard python for loop qrisp would already unroll it on its own.

julfarn:
This PR relates to #76, it almost resolves it except for the required loop unrolling. 

I added a small integration test setup in the form of a shell script. It takes in a qrisp program (required to output MLIR), passes through qcc, and emits the output recording from qir-runner. The external tools are managed in a local python venv using uv. 

We then use this shell script in the RUN line of lit tests (which I take to be the python programs themselves). Lit now also treats `.test.py` extensions as test files. 

Even for a basic bell state qrisp program, we need semantics that were previously not supported: 
Measurement of `!jasp.QubitArray` into an `i64`. This has already been the subject of a discord discussion once. 
This is of course dangerous because it's assumed the qubit array has no more than 64 qubits. 
But I implemented it here so that we don't need to restrict the qrisp subset we support even further. 

Because `qc` only supports bitwise measurement, the single measurement statement needs to be lowered to an `scf.for` loop with bit-fiddling inside. For the moment, the case where the length of the qubit array is known (we require this anyways) is treated separately, and the loop is not constructed, but added in unrolled form. 

I could see removing this distinction and letting the later loop unrolling pass handle this. 
I did not include loop unrolling functionality here to keep things simple. @rainij not sure if we want to wait for upstream changes or use what we added in the loop unrolling branch.


rainij:
>  @rainij not sure if we want to wait for upstream changes or use what we added in the loop unrolling branch.

Let us see if upstream changes are fast enough. If not we copy stuff from the upstream PR and our local PR and make a new one (I would not use the old PR).

Btw: Let us not start discussions from the PR description. It is often used as the merge commit message. Instead you can start the discussion as a normal comment.

mlir/include/qcc/Conversion/ToQIR/Constants.h
```diff
@@ -50,6 +50,11 @@ static constexpr llvm::StringLiteral qirRtResultRecordOutput = "__quantum__rt__r
 /// on the output schema, the label is included in the output or omitted.
 static constexpr llvm::StringLiteral qirRtBoolRecordOutput = "__quantum__rt__bool_record_output";
 
+/// Adds an integer value to the generated output.
+///
+/// Signature: `void(i64,ptr)`.
+static constexpr llvm::StringLiteral qirRtIntRecordOutput = "__quantum__rt__int_record_output";
```
rainij, line 59: This is related to a PR by @ctfle . Right? Might make sense to wait for this PR to be merged first.

mlir/test/tools/qcc/qrisp-integration/integration_run.sh
rainij, line 1: I would organize the testing strategy differently.

- Base the tests on the converted (from py to mlir) tests.
- Still run `qir-runner`, but install via `uv tool install qirrunner` in env.
- Keep the py files as reference, but add a generator script to create/update the actual mlir files from them.
 
This generator script can rely on a venv (with qrisp) but should not be called by the tests. It is only necessary for people actively working on the qrisp stuff. It should also add a comment to the mlir files marking them as generated.

Why the current approach is not good:

- Tests could break on updates or errors in qrisp, thus blocking us.
- The tests slow down considerably, now taking 2000ms instead of just 2ms (for me).
- Adds a dependency to a python package, installs a venv.

Technical issue:
- The `.venv` brings in its own lit tests which are crawled by lit. I found out the hard way by switching to main and getting all these errors on xDSL - thinking WTF!

The qir-runner dependency might still not great but given that one can install it as a standalone tool this might be OK. But let us investigate if its output is stable enough for testing. It might be an option to guard those tests behind a cmake flag which allows failure in certain scenarios. But unsure right now.

I think ideally there should be real integration tests which start from python, but IMO not in this repo. Ideally the responsibility should be shared with qrisp and not entirely our burden. We could turn this into a separate issue.

mlir/test/tools/qcc/qrisp-integration/bell_state.test.py
```diff
@@ -0,0 +1,41 @@
...
+
+    m = measure(qv)
+
+    return m
+
+jaspr = make_jaspr(prepare_bell_state)()
+print(jaspr.to_mlir(lower_stablehlo=True))
+
+
+# A seed of 1 returns integer value 0 (|00> state) for the first shot, and 3 (|11> state) for the second shot, which is consistent with a Bell state.
```
rainij: I would not rely on a seed here. Just check that it is either 0 or 3 via regex. Might still worth pointing out that in-determinism.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -312,30 +313,126 @@ struct ConvertJaspQuantumGateOp final : OpConversionPattern<jasp::QuantumGateOp>
...
 
   LogicalResult matchAndRewrite(jasp::MeasureOp op, OpAdaptor adaptor,
                                 ConversionPatternRewriter& rewriter) const override {
-    auto qcQubit = adaptor.getMeasQ();
+    auto loc = op.getLoc();
+    Value measQOperand = adaptor.getMeasQ();
+    Type measQType = measQOperand.getType();
+    Type resultType = typeConverter->convertType(op.getMeasRes().getType());
+
+    if (isa<MemRefType>(measQType)) {
```
rainij: The nesting of ifs is a bit hard to read (a mixture of deep and long).

I would suggest to perform the very short else branch inside the if first and go on outside the if with the memref case.

You should also consider factoring things out into private methods.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -168,6 +168,23 @@ struct RecordBoolLowering : public OpConversionPattern<aux::RecordBoolOp> {
   }
 };
 
+struct RecordIntLowering : public OpConversionPattern<aux::RecordIntOp> {
```
rainij, line 156: Is this also from the Christoph branch?

mlir/test/tools/qcc-opt/jasp-to-qc-conversion-test.mlir
```diff
@@ -38,3 +39,41 @@ func.func @test() -> tensor<i1> {
     return %random_bit : tensor<i1>
     // CHECK: return [[random_bit]] : i1
 }
+
+// -----
+
```
rainij, line 43: What about testing the other cases? I see only the case of a memref with static size. There is also the single qubit and the dynamic case right?

mlir/test/tools/qcc-opt/jasp-to-qc-conversion-test.mlir
```diff
@@ -38,3 +39,41 @@ func.func @test() -> tensor<i1> {
     return %random_bit : tensor<i1>
     // CHECK: return [[random_bit]] : i1
 }
+
+// -----
```
rainij: What is this for? This file isn't split anyway, right?

rainij (changes requested):

mlir/include/qcc/Conversion/ToQIR/Constants.h
julfarn: I only saw his PR after completing this. We did some things in parallel. Yes, waiting for that PR first is a good idea.

mlir/test/tools/qcc/qrisp-integration/integration_run.sh
julfarn: Not entirely opposed to this, just my view on the criticism:

> Tests could break on updates or errors in qrisp, thus blocking us.

We reference a particular qrisp version. When we want to update the dependency, it suffices to bump the version. Otherwise we'd need to manually rebuild the test files every time. 

> The tests slow down considerably, now taking 2000ms instead of just 2ms (for me).

Only the very first time you run them. After that, the venv has been created and stays cached. We could move it to the build dir though, that would be more appropriate. 

> The .venv brings in its own lit tests which are crawled by lit. I found out the hard way by switching to main and getting all these errors on xDSL - thinking WTF!

This is only an issue if you switch back to a branch before this change, which would never happen after merging this.

As for the dependencies, especially `qir-runner`, I am not so worried. It's easy to disable the tests should they break, and it's a valuable notification to us that a pipeline we intend to support breaks. This doesn't add any ties to our codebase like a C++ library dependency would.

rainij: > Tests could break on updates or errors in qrisp, thus blocking us.
>
> > We reference a particular qrisp version. When we want to update the dependency, it suffices to bump the version. Otherwise we'd need to manually rebuild the test files every time.

Still, this shifts responsibilities. If _our_ tests fail we have to react. But if the error is on qrisp's side those tests would fail too. We cannot know this in advance. Testing the translation from qrisp to mlir should not be our responsibility. It should be done elsewhere.

IMO the tests would be "honest" if we start from the mlir input, because this is what we develop against. The splitting also nicely tracks how qrisp changes. Imagine you re-generate the tests after a qrisp update. If qrisp introduced a change we can nicely point out what the change is and ask for clarification if necessary. With the current ansatz this is less straightforward.

> The tests slow down considerably, now taking 2000ms instead of just 2ms (for me).
>
> > Only the very first time you run them. After that, the venv has been created and stays cached. We could move it to the build dir though, that would be more appropriate.

The two seconds is already after the caching of the venv. In fact, _each_ test takes 2 seconds on its own! With just a few tests you do not see this on a multi-core machine, but try this:

```
LIT_OPTS="--time-tests" cmake --build build/my-dev/ --target test-qcc-project
```

This prints you a nice table of the slowest tests:

```
Slowest Tests:
--------------------------------------------------------------------------
1.98s: QCC_MLIR_COMPILER :: tools/qcc/qrisp-integration/bell_state.test.py
1.97s: QCC_MLIR_COMPILER :: tools/qcc/qrisp-integration/ghz.test.py
0.01s: QCC_MLIR_COMPILER :: tools/qcc/example_ghz.mlir
0.01s: QCC_MLIR_COMPILER :: tools/qcc-opt/ToQIR/export-to-qir-full-pipeline.mlir
0.01s: QCC_MLIR_COMPILER :: tools/qcc-opt/jasp-to-qc-conversion-test.mlir
0.01s: QCC_MLIR_COMPILER :: tools/qcc-opt/jasp-to-qc-gates-test.mlir
...
```

Here you can see that _both_ tests really take that long while all other tests are much faster. _Ordinary_ lit tests should really only take in the order of few milliseconds (slower tests should be the exception, but for these `test.py` it would be intrinsic). I also duplicated the bell test 100 times and ran it with `-j1`, it really takes around 3 minutes (still 18 seconds on 32 cores).

One could now argue that we won't introduce so many tests for qrisp, but I really don't want to needlessly make that bet. Also, the bump from few milliseconds to 2 seconds (almost a factor 1000) is already there today. And that the first run takes even much longer makes it a very bad experience for onboarding.

Please do not take this feedback to personally. I just want to point out something which is I find really important for the quality of the project. Also keep in mind that anything we do now sets an example for future contribution. Somebody else might want to do something similar for other frontends with the justification "you did it already for qrisp".

mlir/test/tools/qcc/qrisp-integration/pyproject.toml
rainij, line 1: For the refactor I propose elsewhere: I believe it might be simpler to just define a `requirements.txt` file (and maybe an additional constraints file) instead of a project file. If we need a similar approach elsewhere the developer can just create a venve somewhere and install the deps as needed. The qirrunner itself can be installed as a "tool" by uv.

mlir/test/tools/qcc/qrisp-integration/integration_run.sh
julfarn: I see your points. I would still like to see some place with a full end-to-end test, but you're right that it would make more sense to be hosted in the qrisp project. 

> Please do not take this feedback to personally.

No worries :) 

> each test takes 2 seconds

Ah yes, I misjudges what was happening here. In hindsight it's obvious, as qrisp -> mlir always takes a noticable time (would be interesting to know the bottleneck: Is it the actual jax tracing that's slow? Or is it the loading of the compiled LLVM into memory?). I agree that this is not acceptable.

I'm going to make the change as you suggested.

rainij: Thanks :) . Please leave this thread open, it is true that a proper integration test is also needed. Normally I would write an issue, but we need to find the right place (definitely not this repo).

julfarn: - split the integration test script into two parts: one for qrisp -> jasp-mlir (to be run manually, outputting mlir test files), and one for the remaining test pipeline (called as part of our tests. I did not measure the timing (for many qubits, qir-runner would become slow here), but they are fast.
- removed `pyproject.toml` and similar files. We have human-readable documentation in the readme for what versions are needed. So I see no advantage in having a `requirements.txt` here. 
- qir-runner is installed via `uv tools` as you suggested. If you ever interact with this, beware that the python package is called `qirrunner`, but the executable is called `qir-runner`. 
- **Additional feature:** The integration tests now also output the qir alongside the qir-runner output. One can filecheck for qir output, output recording, or both.

mlir/test/tools/qcc/qrisp-integration/bell_state.test.py
julfarn: I increased the number of shots and checked for 0-or-3 in each instance. Seed functionality is still there if desired for some test

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: refactored, please have another look

mlir/test/tools/qcc-opt/jasp-to-qc-conversion-test.mlir
julfarn: Tbh I was awaiting feedback before writing too many tests if we decided against this feature. 

I added the dynamic case, the single-qubit case is covered by the pre-existing case at the top of this file.

mlir/test/tools/qcc-opt/jasp-to-qc-conversion-test.mlir
julfarn: Removed.

mlir/test/tools/qcc/qrisp-integration/pyproject.toml
julfarn: As pointed out above, I see no point in the `requirements.txt` because we already have the README. There's only two items in there anyways. If it gets more complicated, we can add it. But then the README should reference it. In any case I'd wait for a qrisp release.

mlir/include/qcc/Conversion/ToQIR/Constants.h
julfarn: Now using Christoph's version.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
julfarn: Now using his version, this file is no longer in the diff.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -314,31 +315,144 @@ struct ConvertJaspQuantumGateOp final : OpConversionPattern<jasp::QuantumGateOp>
...
+/// Example (single qubit):
 /// ```mlir
 /// %measured, %state1 = jasp.measure %q, %state0 : !jasp.Qubit, !jasp.QuantumState -> tensor<i1>, !jasp.QuantumState
 /// ```
 /// becomes:
 /// ```mlir
 /// %measured = qc.measure %q : !qc.qubit -> i1
 /// ```
+///
+/// Example (qubit array, returning tensor<i64>):
```
rainij: You could just explain verbally what it does. This many lines of non-syntax hightlighted mlir is a bit hard to read and I fear that it might become outdated. Now that this is tested we can rely on the tests to provide verbose examples.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
```diff
@@ -314,31 +315,144 @@ struct ConvertJaspQuantumGateOp final : OpConversionPattern<jasp::QuantumGateOp>
...
+    auto loc = op.getLoc();
+    Value measQOperand = adaptor.getMeasQ();
+    Type measQType = measQOperand.getType();
+
+    // single qubit measurement
+    if (!isa<MemRefType>(measQType)) {
+      return convertSingleQubitMeasurement(op, rewriter, loc, measQOperand);
+    }
+
+    // TODO: Once loop unrolling is in the pipeline, we can eliminate the static case here.
```
rainij, line 345: Created an issue for that, so that we do not miss it: https://github.com/FullStaQD/compiler/issues/86 .

mlir/test/tools/qcc/qrisp-integration/qrisp_sources/bell_state.py
```diff
@@ -0,0 +1,63 @@
+#  ===----------------------------------------------------------------------===//
```
rainij: Is this license header enforced by tooling? IMO this is overkill. I would like to drop it.

mlir/test/tools/qcc/qrisp-integration/qrisp_sources/bell_state.py
rainij, line 1: I would not introduce another folder here. Just leave it side by side with the generated tests.

mlir/test/tools/qcc/qrisp-integration/qrisp_sources/bell_state.py
```diff
@@ -0,0 +1,63 @@
...
+    m = measure(qv)
+
+    return m
+
+mlir = str(make_jaspr(prepare_bell_state)().to_mlir(lower_stablehlo=True))
+print(
+f"""
+// RUN: %python %S/integration_run.py -n 5 -c qcc  %s | FileCheck %s
+
+{mlir}
```
rainij: IMO adding the checks here is too much responsibility for this script. Whenever the mlir is generated I expect the developer to carefully go over it and add the CHECKs the usual way (possibly using the standard generator script provided by LLVM).

The only things this script should really add is something like:

```
// some comment that this was generated

{mlir}
```

Maybe it makes sense to add the RUN line, but I would directly call `qcc`, not a python script. But let us discuss this detail elsewhere.

mlir/test/tools/qcc/qrisp-integration/qrisp_generate_mlir.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+which python3 >/dev/null 2>&1 || { echo "python not found. Please activate a venv with qrisp and xDSL installed." >&2; exit 1; }
+python3 -c "import qrisp" >/dev/null 2>&1 || { echo "qrisp not found. Please activate a venv with qrisp installed." >&2; exit 1; }
+python3 -c "import xdsl" >/dev/null 2>&1 || { echo "xDSL not found. Please activate a venv with xDSL installed." >&2; exit 1; }
```
rainij: I think the check for `qrisp` is OK, but `xdsl` is too much of a detail IMO.

mlir/test/tools/qcc/qrisp-integration/qrisp_generate_mlir.sh
```diff
@@ -0,0 +1,20 @@
...
+
+
+# Run all python scripts in the qrisp_sources directory, storing the results in the current directory.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
rainij: Running literally anything is not a good practice from a security perspective. Even leaving this aside it is not great to reason about such a script. Basically this script here says: Do whatever the other scripts say. Our intention is a bit narrower: For each qrisp program in this folder generate the corresponding mlir (for testing).

mlir/test/tools/qcc/qrisp-integration/integration_run.py
rainij, line 1: This is a very complicated script for the simple thing we want. I believe we do not need this.

mlir/test/tools/qcc/qrisp/ghz_from_qrisp.mlir
rainij, line 1: I think the setup is still too complicated and it is not really clear what exactly is accomplished by the scripts. I think it automates too much. For example I wonder what exactly is the workflow if a CHECK fails after generating? Now there are two places where I can find the checks. The fastest way to verify I updated it correctly is to put into the mlir files and see if tests succed. On the other hand I have to remember to also update the python file.

I suggest a more manual approach. As said elsewhere I consider a test update something where a developer checks each test which is updated _individually_. Having this in mind, there is not much gain in a script which updates everything at once.

IMO the workflow of updating tests should be like this:
- Developer decides to update `bell.py`.
- runs `python bell.py` to obtain new mlir.
- copies the new mlir into `bell.generated.mlir` and checks the diff carefull.
- If necessary updates the CHECKs as you would do with any other test updated - nothing special here. Use LLVM's generator script if necessary.
- If anything looks fishy and might indicate an inconsistency between qrisp and qcc raise the issue somewhere and do not just go on.

IMO one could indeed replace `python bell.py` by a thin wrapper which imports the entry point and standardizes the output, e.g. by printing a small comment that this is generated first and then the mlir. This way the qrisp programs stay lean and can focus on just defining the entry point. I could imagine that the comment should also print the qrisp version being used.

The setup could be documented in this README (I would keep it). I expect it to be very brief.

mlir/test/tools/qcc/qrisp-integration/bell_state.generated_test.mlir
```diff
@@ -0,0 +1,49 @@
+
+// RUN: %python %S/integration_run.py -n 5 -c qcc  %s | FileCheck %s
```
rainij: This should run `qcc` directly. Also, you have effectively two tests here: one with QIR output, the other with qir-runner. Standard practice is to make this explicit by providing _two_ RUN lines. See [FileCheck -check-prefixes](https://llvm.org/docs/CommandGuide/FileCheck.html#the-filecheck-check-prefixes-option). Something like (or similar):

// RUN: qcc %s | FileCheck %s
// RUN: qcc %s | qir-runner | FileCheck %s -check-prefixes=CHECK-RUNNER

mlir/test/lit.cfg.py
```diff
@@ -59,7 +59,7 @@ class ConfigType:
 llvm_config.use_default_substitutions()
 
 # excludes: A list of directories and filenames to exclude from the testsuite.
-config.excludes = []
+config.excludes = [".venv"]
```
rainij: I would not do this. That protects only users who put their venv into the test folder (which they shouldn't) and call it `.venv` (personal preference).

rainij (changes requested):
I would like to simplify the setup even further. Please have a look.

mlir/test/tools/qcc/qrisp-integration/pyproject.toml
rainij: I believe it makes sense to put it into a requirements.txt. This way the README can reference it and my command to install the stuff stays the same. Even for one item I would do it. This way nothing changes in case it gets more complicated. On the other hand I see really no down-sides.

Other thing: you can also add comments to the requirements file, like `xdsl` not needed explicitly in the future, see this or that issue.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
rainij: Better now.

rainij:
In hindsight it might have been better to not put two topics into this PR:

- Update of the jasp to qir pass
- Update of the testing infra for qrisp

I think the latter will block the former to be merged for some time. And even the other way around: the former blocked the latter for some time because you waited for stuff by @ctfle .

mlir/test/tools/qcc/qrisp-integration/pyproject.toml
julfarn: added requirements.txt and updated readme, please have a look

mlir/test/tools/qcc/qrisp-integration/qrisp_sources/bell_state.py
julfarn: Yes, unfortunately it is enforced. We could change that, but I'd prefer not to in this PR.

mlir/test/tools/qcc/qrisp-integration/qrisp_sources/bell_state.py
julfarn: changed

mlir/test/tools/qcc/qrisp-integration/qrisp_generate_mlir.sh
julfarn: removed this script

mlir/test/tools/qcc/qrisp-integration/integration_run.py
julfarn: removed this script

mlir/test/lit.cfg.py
julfarn: changed

mlir/test/tools/qcc/qrisp-integration/bell_state.generated_test.mlir
julfarn: It's a little more complicated unfortunately, but I made the change. The integration script is no longer needed. But if this pipeline changes, we have to change every such RUN line.

mlir/test/tools/qcc/qrisp/ghz_from_qrisp.mlir
julfarn: I eliminated the automation completely. Only the python scripts outputting pure MLIR remain.

mlir/test/tools/qcc/qrisp-integration/qrisp_sources/bell_state.py
julfarn: Personally I really like having a single source of truth for a test case (meaning you're not allowed to touch the `.generated.mlir`, you have to change the `.py` and regenerate. 
Having the actual program input side by side with the actual output in one file was my main motivation.

But I can see where you're coming from, and I made the change.

mlir/lib/Conversion/JaspToQC/JaspToQC.cpp
julfarn: Yes, the explanation a few lines above is sufficient. removed.

mlir/test/tools/qcc/qrisp-integration/qrisp_generate_mlir.sh
julfarn: Removed this file

julfarn:
I removed the automation completely as you suggested. Please have a look at the new README file. 

One last comment on performance: The integration test checking for QIR and qirrunner outputs (two tests, really) now takes 0.17s , where other integration tests take 0.05 and simple single-pass tests take 0.01 or 0.02. I believe this is an OK overhead for using qirrunner, especially since this is parallelized.

mlir/test/tools/qcc/qrisp-integration/README.md
```diff
@@ -0,0 +1,72 @@
+# From Qrisp to MLIR: Step-by-Step Guide
+
+This guide documents the process of converting a [Qrisp](https://qrisp.eu/) quantum program into an MLIR test case for `qcc`.
+
+---
+
+## Prerequisites
+
+This project requires a Python environment set up by the `requirements.txt` in this folder. You can install it via pip:
```
rainij:
```suggestion
Setup a python environment and do:
```

mlir/test/tools/qcc/qrisp-integration/README.md
```diff
@@ -0,0 +1,72 @@
...
+    cx(qv[0], qv[1])
+    mes0 = measure(qv[0])
+    mes1 = measure(qv[1])
+    return
+
+jaspr = make_jaspr(bell)()
+print(jaspr.to_mlir(lower_stablehlo=True))
+```
+
+This example creates a bell state state and measures all the qubits.
```
rainij:
```suggestion
```

mlir/test/tools/qcc/qrisp-integration/README.md
```diff
@@ -0,0 +1,72 @@
...
+Run the script:
+
+```bash
+python my_program.test.py > my_program.generated_test.mlir
```
rainij: Why "generated_test" instead of "generated"? Our other tests almost never end in "test.mlir".

mlir/test/tools/qcc/qrisp-integration/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```
+// Test 1: Check QIR output.
+// RUN: qcc  %s | mlir-translate -mlir-to-llvmir | FileCheck %s --check-prefix=CHECK-QIR
+// Test 2: Check output recording from qir-runner simulation.
+// qir-runner requires a .ll file (it checks the extension and does not accept
+// stdin), so we first emit the LLVM IR to a temporary .ll file and then invoke
+// qir-runner on it.
+// RUN: qcc %s | mlir-translate -mlir-to-llvmir > %t.ll
+// RUN: uvx --from qirrunner qir-runner --file %t.ll -s 5 | FileCheck %s --check-prefix=CHECK-SIM
```
rainij: Here I would like to avoid depending on `uvx`. The tests should directly call `qir-runner`. It shouldn't matter how it is provisioned: uv, apt, directly downloaded, build from source etc.

mlir/test/tools/qcc/qrisp-integration/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+---
+
+## Step 3 - Add FileCheck directives
```
rainij: IMO step 2 and 3 could be made more concise. They explain too much detail, some of it generic FileCheck stuff.

You could emphasise a few best practices instead. For example you could mention that in order to have fast tests the programs should be as minimal as possible (little qubit count probably matters most).

mlir/test/tools/qcc/qrisp-integration/qrisp_sources/bell_state.py
rainij: Can you open an issue and ask relevant people for feedback (Daniel, Lukas)?

mlir/test/tools/qcc/qrisp-integration/bell_state.py
rainij, line 1: I would call the two files just `bell_state.py` and `bell_state.mlir` (without "test" and "generated"). Instead I would just add a comment to the file that it is generated. That is common practice in LLVM. The README already indicates that most tests are generated and it follows the philosophy that each test has its identity. That they are partly generated is not the most important aspect when viewing the folder. We have also other tests which are generated in part (the CHECK lines).

mlir/test/tools/qcc/qrisp-integration/bell_state.test.py
```diff
@@ -0,0 +1,31 @@
...
+    return m
+
+mlir = str(make_jaspr(prepare_bell_state)().to_mlir(lower_stablehlo=True))
+print(
+f"""
+// GENERATED FROM QRISP.
+
+{mlir}
+"""
+)
```
rainij: This part here would be nice to be outsourced and standardized. We could add a folder `/mlir/utils/` and under that one a small script `generate_qrisp_mlir.py` which imports this file dynamically and imports the function `entry_point` (assuming that the name is standardized).

I would also like to generate a comment like this:

```
// GENERATED FROM QRISP VERSION 0.55.5
```

I know right now we have a commit, but there is like something one can do. I think it is important to document to _which_ qrisp the mlir belongs.

mlir/test/tools/qcc/qrisp-integration/README.md
rainij: At least in the devcontainer you can make it easily available like so:

```docker
RUN uv tool install cmake && \
  uv tool install ninja && \
  uv tool install lit && \
  uv tool install prek && \
  uv tool install qirrunner
```

I use it locally exactly like this. I do not know if our CI setup easily supports it. But it should. If not, raise an issue and we find a temporary solution.

mlir/test/tools/qcc/qrisp-integration/bell_state.generated_test.mlir
```diff
@@ -0,0 +1,52 @@
+// Test 1: Check QIR output.
+// RUN: qcc  %s | mlir-translate -mlir-to-llvmir | FileCheck %s --check-prefix=CHECK-QIR
+// Test 2: Check output recording from qir-runner simulation.
+// qir-runner requires a .ll file (it checks the extension and does not accept
+// stdin), so we first emit the LLVM IR to a temporary .ll file and then invoke
+// qir-runner on it.
+// RUN: qcc %s | mlir-translate -mlir-to-llvmir > %t.ll
+// RUN: uvx --from qirrunner qir-runner --file %t.ll -s 5 | FileCheck %s --check-prefix=CHECK-SIM
```
rainij:
```suggestion
// RUN: qcc %s | mlir-translate -mlir-to-llvmir > %t.ll
// RUN: FileCheck %s --check-prefix=CHECK-QIR < %t.ll
// RUN: qir-runner --file %t.ll -s 5 | FileCheck %s --check-prefix=CHECK-SIM
```

It is evident that this is two tests so no need for those comments. I would also leave out the other comment, I do not want to see it repeated over and over again in multiple files, or discussions in PRs (you forgot to add this comment ...).

rainij (changes requested):

mlir/test/tools/qcc/qrisp-integration/qrisp_sources/bell_state.py
julfarn: See #89

mlir/test/tools/qcc/qrisp-integration/README.md
julfarn: changed, also in existing filenames

mlir/test/tools/qcc/qrisp-integration/README.md
julfarn: changed

mlir/test/tools/qcc/qrisp-integration/README.md
julfarn: Agreed. Have another look please.

mlir/test/tools/qcc/qrisp-integration/bell_state.py
julfarn: True, it's also just incorrect as they are no longer generated. Changed.

mlir/test/tools/qcc/qrisp-integration/bell_state.test.py
julfarn: Added this generator script. It has to call `uv pip freeze`, I didn't find a way to get the (correct) version by other means.

mlir/test/tools/qcc/qrisp-integration/bell_state.generated_test.mlir
julfarn: Fair enough, removed.

mlir/test/tools/qcc/qrisp-integration/README.md
julfarn: Changed to 
```
Setup a python environment and run:
```

mlir/test/tools/qcc/qrisp-integration/README.md
julfarn: Could we keep uvx as the temporary solution? I'd create an issue and ask the relevant people to have it installed on the CI runner.

rainij: That is OK, add the issue to milestone 0.1.0 and make it high prio. But already add it to the devcontainer. Add a TODO to the RUN line (referencing the issue).

mlir/test/tools/qcc/qrisp-integration/bell_state.mlir
```diff
@@ -0,0 +1,49 @@
+// RUN: qcc  %s | mlir-translate -mlir-to-llvmir | FileCheck %s --check-prefix=CHECK-QIR
+// RUN: qcc %s | mlir-translate -mlir-to-llvmir > %t.ll
+// RUN: uvx --from qirrunner qir-runner --file %t.ll -s 5 | FileCheck %s --check-prefix=CHECK-SIM
```
rainij:
```suggestion
// RUN: qcc %s | mlir-translate -mlir-to-llvmir > %t.ll
// RUN: FileCheck %s --check-prefix=CHECK-QIR < %t.ll
// RUN: uvx --from qirrunner qir-runner --file %t.ll -s 5 | FileCheck %s --check-prefix=CHECK-SIM
```

IMO this is nicer. Generating the tmp file once and then use it everywhere.

julfarn: Not only that, but a performance improvement 🧠

mlir/test/tools/qcc/qrisp-integration/README.md
julfarn: Done, see #92

mlir/test/tools/qcc/qrisp-integration/__pycache__/bell_state.cpython-314.pyc
rainij, line 1: Should not be checked in.

mlir/test/tools/qcc/qrisp-integration/bell_state.py
```diff
@@ -0,0 +1,21 @@
+#  ===----------------------------------------------------------------------===//
```
rainij, line 1: Reminder to myself: decide what to do about this license header before merge. See also https://github.com/FullStaQD/compiler/issues/89 .

mlir/utils/generate_qrisp_mlir.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python3
+
+#  ===----------------------------------------------------------------------===//
```
rainij, line 3: In this file it _might_ be somewhat OK to have a license header. But I would leave it out if not necessary.

mlir/utils/generate_qrisp_mlir.py
rainij, line 1: Not the most beautiful script I have ever seen. Would be nice if you could inspect it for opportunities of improvement. But I leave it to your own judgement. Please resolve once you are satisfied.

rainij (changes requested):
Almost good to go.

mlir/test/tools/qcc/qrisp-integration/__pycache__/bell_state.cpython-314.pyc
julfarn: oops, sorry

mlir/utils/generate_qrisp_mlir.py
julfarn: It doesn't look nice but I think the functionality of looking for an arbitrary single function in the source is convenient. 

I changed a few things. When we move to released versions of qrisp we can simplify the version string retrieval.

Your comment is of course that this is vibe coded, and one can tell. But I think this is a case where we could embrace that the code does not have to be of the highest quality, as long as it works, and is managed more by machines that humans. (I went over it of course, and already simplified a few things already.)

rainij: I actually disagree, but it is currently a low priority thing. Once we update it we might want to refactor it though. Please resolve after reading.

rainij (approved):

julfarn:
@burgholzer, would you like to have a look?

mlir/utils/generate_qrisp_mlir.py
burgholzer, line 1: This can be made more convenient.
If you add inline script metadata
```Python
# /// script
# dependencies = [
#   "qrisp @ git+https://github.com/eclipse-qrisp/Qrisp.git@b81ea2f979d21cd8d600e79d8b0c7066fe7cbe1b",
#   "xdsl==0.59.0"
# ]
# ///
```
(plus the comments)
and replace the shebang line with `#!/usr/bin/env -S uv run --script --quiet`, then generating a new test case is as easy as

```bash
./generate_qrisp_mlir.py my_program.py > my_program.mlir
```

No need to set up a virtual environment; no need to manually install dependencies. Just a single line.
(The script already depends on the presence of `uv`; so the change in the shebang line is presumably fine)

See, for example, https://github.com/munich-quantum-toolkit/core/blob/main/noxfile.py or https://github.com/munich-quantum-toolkit/core/blob/main/eval/dd_evaluation.py for how this us used in MQT Core in practice.

Also puts everything related to the conversion in one place.

The documentation in the README should be adjusted accordingly.

burgholzer (changes requested):
This generally looks good to me. I just have one suggestion on improving the ergonomics of the test generation taking advantage of some modern tooling and standards.
Otherwise, this seems ready to go in.

mlir/utils/generate_qrisp_mlir.py
julfarn: Wow, thanks. I made the changes and deleted the requirements.txt. 

I did not know about inline script metadata until you mentioned it this morning, seems very powerful!

burgholzer: It's a pretty neat technique for shipping self contained python scripts and pairs really well with uv. 🙂

burgholzer (approved):
LGTM 👍🏻

rainij:
@julfarn can you have a look what exactly is blocking the merge? Github claims there is still an open conversation. I suspect it is one of those where the resolve button is missing. I cannot investigate right now.

PR #83, opened 2026-06-15
Add .clangd to gitignore

julfarn:
git-ignore clangd settings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /build
 /CMakeUserPresets.json
+/.clangd
 
 # vscode
 /.vscode


julfarn:
Not sure if this makes sense. My .clangd has some settings specific to my setup. We could include it later if we decide to have a project-wide .clangd.

rainij:
> Not sure if this makes sense. My .clangd has some settings specific to my setup. We could include it later if we decide to have a project-wide .clangd.

No it is actually good to gitignore it. I think it should be a personal thing. LLVM also gitignores it I believe.

rainij (approved):

rainij:
@julfarn please merge yourself. There is an email in the one of the commit message you might not want to share (the squash automatism of github found it). (You can just remove the corresponding line from the generated message if this is what you want.)

PR #84, opened 2026-06-15
Update devcontainer to use llvm 22.1.7

rainij:


rainij:
@burgholzer I hope this is correct. I do not use the default devcontainer.

burgholzer (approved):
LGTM

PR #88, opened 2026-06-18
Preliminary loop unrolling

### LINKED ISSUE #46: Loop unrolling
We want to support loop unrolling for backands uncapable of (backward) control flow. In the first versions, loop unrolling will be mandatory. 
Going forward, we will gradually increase support for loops throughout the whole pipeline, considering vectorization as well.

- [ ]  Add loop unrolling to the pipeline.
- [ ]  Ensure that all loops have been unrolled, otherwise emit error. In particular, all loops must have a static number of iterations.
- [ ]  Implement test cases.

### LINKED ISSUE #96: GHZ test case
Add a GHZ test case to the qrisp integration tests. Exercises for loops.

julfarn:
Resolves #46. Resolves #96. 

Added passes while-to-for and tmp-raise-scf-to-affine. Extended compilation pipeline.


mlir/include/qcc/Conversion/AffineRaise/AffineRaise.td
```diff
@@ -0,0 +1,63 @@
...
+  ];
+}
+
+
+def WhileToFor : Pass<"while-to-for"> {
+  let summary = "Convert scf.while loops to scf.for loops using uplift patterns.";
+
+  let description = [{
+      Applies the SCF uplift patterns to rewrite eligible `scf.while` loops as
+      `scf.for` loops.
```
rainij: You might want to explicitly point out that those patterns are builtin (into MLIR).

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: qcc-opt %s -while-to-for | FileCheck %s
+
+func.func public @main() -> i64 {
```
rainij: find a better "test name".

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
```diff
@@ -0,0 +1,27 @@
...
+  %step = arith.constant 1 : i64
+  %lb = arith.constant 1 : i64
+  %zero = arith.constant 0 : i64
+  %ub_incl = arith.constant 4 : i64
+  %loop_out_iv, %loop_out_sum = scf.while (%induction = %lb, %sum = %zero) : (i64, i64) -> (i64, i64) {
+    %continue_condition = arith.cmpi sle, %induction, %ub_incl : i64
+    scf.condition(%continue_condition) %induction, %sum : i64, i64
+  } do {
```
rainij: Some of the variable names are a bit cumbersome. Maybe:

- "induction" to "iv" or just "i"
- "loop_out_iv" to "i_out"
- "loop_out_sum" to just "sum".
- "continue_condition" to "done" (this name is really bad)

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: qcc-opt %s -while-to-for | FileCheck %s
+
+func.func public @main() -> i64 {
+  %step = arith.constant 1 : i64
+  %lb = arith.constant 1 : i64
+  %zero = arith.constant 0 : i64
+  %ub_incl = arith.constant 4 : i64
+  %loop_out_iv, %loop_out_sum = scf.while (%induction = %lb, %sum = %zero) : (i64, i64) -> (i64, i64) {
```
rainij: Does it actually matter for the pass if the IV is returned? Normally I would ask for a test of both cases, but this isn't something you implemented so it is OK that it is not tested. Just curious.

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
rainij, line 1: We should indicate that we only test "our" part of the implementation. Otherwise one might wonder why there isn't much testing going on.

IMO there are two things we might want to test:
- The sle thing you implemented specifically (this is a must)
- _Simple_ example demonstrating how it generally works - just in case upstream changes it and for reference. (Maybe optional, but I would do it)

If we have a test for the second thing we can simplify the test for the first thing. (easier example, less verbose CHECK lines, focusing on the actual thing under test).

mlir/test/tools/qcc/simple_loop.mlir
rainij, line 1: I would like to wait for your other PR on the qrisp testing infra to be merged. Then you could replace this test by a generated one which has a loop in it.

Just in case you have it here to explicitly test the unrolling: IMO this is just what we do right now. More important is that we can "do something" if loops are present.

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -67,8 +67,32 @@ void buildQuantumPipeline(mlir::PassManager& pm) {
...
+  // they can subsequently be raised to `affine.for`.
+  pm.addNestedPass<mlir::func::FuncOp>(qcc::createWhileToFor());
+
+  // Raise `scf.for` loops to `affine.for` where the bounds and step permit it.
+  pm.addNestedPass<mlir::func::FuncOp>(qcc::createTmpRaiseSCFToAffinePass());
+
+  // Unroll affine loops.
+  // TODO: We have to do this in this by parsing because the createAffineLoopUnroll function does not pass on the -1
+  // factor.
+  mlir::affine::registerAffineLoopUnroll();
```
rainij, line 86: Shouldn't this be part of `qcc.cpp` tool? Where the other register functions live.

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -67,8 +67,32 @@ void buildQuantumPipeline(mlir::PassManager& pm) {
...
+  if (failed(mlir::parsePassPipeline("func.func(affine-loop-unroll{unroll-factor=-1})", pm))) {
+    llvm::errs() << "Failed to parse pass pipeline\n";
+    return;
```
rainij: This returns a half built pipeline upon error. Abort here execution in some way (figure out what is best: assert, unreachable, exit, etc).

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -67,8 +67,32 @@ void buildQuantumPipeline(mlir::PassManager& pm) {
...
+
+  // Convert `scf.while` loops into `scf.for` loops where possible, so that
+  // they can subsequently be raised to `affine.for`.
+  pm.addNestedPass<mlir::func::FuncOp>(qcc::createWhileToFor());
+
+  // Raise `scf.for` loops to `affine.for` where the bounds and step permit it.
+  pm.addNestedPass<mlir::func::FuncOp>(qcc::createTmpRaiseSCFToAffinePass());
+
+  // Unroll affine loops.
+  // TODO: We have to do this in this by parsing because the createAffineLoopUnroll function does not pass on the -1
```
rainij, line 82: Might makes sense to mention that -1 gets translated to 4. And that we consider this a bug upstream (and that you are on it).

mlir/lib/Conversion/AffineRaise/WhileToFor.cpp
```diff
@@ -0,0 +1,92 @@
...
+    (void)applyPatternsGreedily(getOperation(), std::move(patterns));
+  }
+};
+} // namespace
+
+} // namespace qcc
+
+namespace impl {
+static std::unique_ptr<::mlir::Pass> createWhileToFor() { return std::make_unique<qcc::WhileToFor>(); }
+} // namespace impl
```
rainij: This is likely dead code. Tablegen generates a function with that name in a different namespace.

mlir/tools/qcc-opt/qcc-opt.cpp
```diff
@@ -46,15 +50,18 @@ int main(int argc, char** argv) {
...
       // clang-format off
+    mlir::affine::AffineDialect,
     mlir::func::FuncDialect,
     mlir::arith::ArithDialect,
     mlir::tensor::TensorDialect,
     mlir::bufferization::BufferizationDialect,
     mlir::linalg::LinalgDialect,
     mlir::cf::ControlFlowDialect,
     mlir::scf::SCFDialect,
+    mlir::memref::MemRefDialect,
```
rainij, line 62:
```suggestion
```

Duplicate

mlir/include/qcc/Conversion/AffineRaise/AffineRaise.td
```diff
@@ -0,0 +1,63 @@
...
+    "mlir::affine::AffineDialect",
+    "mlir::arith::ArithDialect",
+  ];
+}
+
+
+def WhileToFor : Pass<"while-to-for"> {
+  let summary = "Convert scf.while loops to scf.for loops using uplift patterns.";
+
+  let description = [{
```
rainij, line 55: Could you briefly explain what it does? Currently one has to rely on the name to figure it out. IMO it makes sense to explain it in some way, either words or via examples, whatever is best. In particular mention that sle and slt are both supported.

I would also appreciate a few words on undefined behavior. Similar to the discussion we had on discourse. I think there are a few corner cases which might behave surprising. If it is hard to pin down explain it at least roughly so that everybody is aware of the issue. For example for the SLE to SLT thing you add 1 somewhere. I believe this could overflow. IMO no need to change, but we should document this.

mlir/lib/Compiler/Pipeline.cpp
```diff
@@ -67,8 +67,32 @@ void buildQuantumPipeline(mlir::PassManager& pm) {
...
+  // factor.
+  mlir::affine::registerAffineLoopUnroll();
+  if (failed(mlir::parsePassPipeline("func.func(affine-loop-unroll{unroll-factor=-1})", pm))) {
```
rainij: What about nested loops? Like this one

```mlir
  func.func @nest(%m: memref<2x2xf32>, %v: f32) {
    affine.for %i = 0 to 2 {
      affine.for %j = 0 to 2 {
        affine.store %v, %m[%i, %j] : memref<2x2xf32>
      }
    }
    return
  }
```

An easy fix for this example would be to add `unroll-num-reps=2` or taking a "sufficiently high" value (it looks like -1 does not work here). A proper fix would be to apply until convergence, but I would be fine with a "reasonably high" value and adding a comment about it.

mlir/include/qcc/Conversion/AffineRaise/AffineRaise.td
julfarn: changed

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
julfarn: changed

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
julfarn: Not sure. I feel like it has to be returned, because it needs to be `scf.yield`ed into the next iteration, so it is automatically returned in the last iteration. 

There's extra arithmetic to recover the (final) induction var if it is **used**, though. (We went over this once.)

mlir/lib/Compiler/Pipeline.cpp
julfarn: In qcc, we don't have to register passes. This applies only to qcc-opt, where we need to be able to parse pipelines. 

The registration is only needed for as long as we need to keep the parsing workaround.

mlir/lib/Compiler/Pipeline.cpp
julfarn: True, I went with unreachable.

mlir/lib/Compiler/Pipeline.cpp
julfarn: changed

mlir/lib/Conversion/AffineRaise/WhileToFor.cpp
julfarn: good catch!

mlir/tools/qcc-opt/qcc-opt.cpp
julfarn: removed

mlir/lib/Compiler/Pipeline.cpp
julfarn: hm, I see. I have to look into this again. The unroll-num-reps argument is also only available from pipeline parsing. I don't think this pass is well maintained.

Set to 10 for now.

mlir/include/qcc/Conversion/AffineRaise/AffineRaise.td
julfarn: I added an example and extended the description a bit. 

I thought a bit about undefined behavior, but I didn't go through the entire uplift-while-to-for pattern. I don't immediately see that something could be going wrong, except for the +1 in SLE to SLT. For that I added a small remark.

If this becomes relevant for us (maybe we are forced by hardware to deal with very narrow integer types), we should examine the pattern more closely.

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
julfarn: split up the tests as you suggest

mlir/test/tools/qcc/simple_loop.mlir
julfarn: Deleted this file and added checks to `ghz.mlir`.

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
rainij: I also vaguely remember this, but I do not really remember what the issue was. Maybe we were wrong. I believe that the while loop can return a different type than what yield returns. Only `scf.condition` has to be in sync with the return value. But I do not have the details in my head. Can you have a look into the spec of scf.while again?

I leave it to you how to proceed with this thread (including closing).

mlir/lib/Compiler/Pipeline.cpp
rainij: Can you add a super simple test case (like the above with just two reps)? CHECK lines should only make sure that the loops are gone.

rainij (changes requested):
Almost good to go, just tiny stuff left.

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
julfarn: From the docs: 

> The op expects the operand types to match with argument types of the “before” region; the result types to match with the trailing operand types of the terminator of the “before” region, and with the argument types of the “after” region.

Somehow the information of the induction variable needs to be passed along. One could in theory store it to an outer-scope memref, then pass nothing from the before to the after region (so that nothing will be returned), and re-load it from the memref in the after region to feed it back into the before region. 

But I don't think we have to consider a contrived case like this. The uplifting pattern will most likely not match on it.

mlir/lib/Compiler/Pipeline.cpp
julfarn: added a test. There's also a bug #99 I found on the way.

At first I thought to test the single pass in `qcc-opt`, but it doesn't really make sense because 
- we would have to manually specify the pass arguments, thus not reliably testing the actual pass as used in `qcc`
- The pass is of course tested in mlir. 

That's why I went with a test for `qcc`.

mlir/test/tools/qcc-opt/AffineRaise/while-to-for.mlir
rainij: Thanks, makes sense. I believe that was what we came up with last time. Hope I won't forget it until next time :D.

mlir/lib/Compiler/Pipeline.cpp
rainij: Makes sense. Thanks.

rainij (approved):

rainij:
@burgholzer can this be merged?

mlir/lib/Conversion/AffineRaise/SCFToAffine.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+  if (auto lbMaxOp = lb.getDefiningOp<affine::AffineMaxOp>()) {
+    lbOperands = lbMaxOp->getOperands();
+    lbMap = lbMaxOp.getAffineMap();
+  }
+
+  auto affineFor =
+      affine::AffineForOp::create(rewriter, op.getLoc(), lbOperands, lbMap, ubOperands, ubMap, step, op.getInits());
+
+  return std::make_pair(affineFor, affineFor.getInductionVar());
```
burgholzer:
```suggestion
  return {affineFor, affineFor.getInductionVar()};
```
less verbose syntax 🤷🏼

mlir/lib/Conversion/AffineRaise/SCFToAffine.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+  // old_iv = old_lb + new_iv * step
+  AffineMap ivMap = AffineMap::get(2, 1, d0 + d1 * s0);
+
+  llvm::SmallVector<Value, 3> ivOperands = {lb, affineFor.getInductionVar(), step};
+
+  rewriter.setInsertionPointToStart(affineFor.getBody());
+  auto oldIV = affine::AffineApplyOp::create(rewriter, op.getLoc(), ivMap, ivOperands);
+
+  return std::make_pair(affineFor, oldIV);
```
burgholzer:
```suggestion
  return {affineFor, oldIV};
```

mlir/lib/Conversion/AffineRaise/SCFToAffine.cpp
```diff
@@ -0,0 +1,358 @@
...
+//===----------------------------------------------------------------------===//
+// Pass implementation
+//===----------------------------------------------------------------------===//
+
+void TmpSCFToAffine::runOnOperation() {
+  MLIRContext& ctx = getContext();
+  RewritePatternSet patterns(&ctx);
+  populateSCFToAffineConversionPatterns(patterns);
+
+  (void)applyPatternsGreedily(getOperation(), std::move(patterns));
```
burgholzer: Not a big fan of void casts (or C-style casts in general).
Either the pattern application should be explicitly checked (like here https://github.com/munich-quantum-toolkit/core/blob/cad617b53eb8e191d85bde5e058cfd2f445a933a/mlir/lib/Dialect/QCO/Transforms/Optimizations/MergeSingleQubitRotationGates.cpp#L677-L679) or the C++ construct `std::ignore` shall be used in my opinion.
Slight preference for the first solution.

mlir/lib/Conversion/AffineRaise/WhileToFor.cpp
```diff
@@ -0,0 +1,88 @@
...
+namespace {
+struct WhileToFor final : public impl::WhileToForBase<WhileToFor> {
+  using impl::WhileToForBase<WhileToFor>::WhileToForBase;
+
+protected:
+  void runOnOperation() override {
+    RewritePatternSet patterns(&getContext());
+    patterns.add<ConvertWhileSLEToSLT>(&getContext());
+    mlir::scf::populateUpliftWhileToForPatterns(patterns);
+    (void)applyPatternsGreedily(getOperation(), std::move(patterns));
```
burgholzer: Same comment as in the other file applies here.

burgholzer (changes requested):
Just some minor nits. Otherwise this LGTM.

I do not want to make a big discussion out of this, but I think one could discuss about the naming of the pass and where it is currently placed. I thought this is exactly the kind of thing the `aux` dialect is for. Yet, this is not placed anywhere near that and is also called `Tmp` instead of `aux`. Feels slightly inconsistent to the inner monk. But in the end, I do not care that much. If all the other points are addressed, this can be merged from my side.

rainij:
> I do not want to make a big discussion out of this, but I think one could discuss about the naming of the pass and where it is currently placed. I thought this is exactly the kind of thing the `aux` dialect is for. Yet, this is not placed anywhere near that and is also called `Tmp` instead of `aux`. Feels slightly inconsistent to the inner monk. But in the end, I do not care that much. If all the other points are addressed, this can be merged from my side.

Actually I have a similar feeling. On the other hand I want to merge it ASAP. We can probably defer it a little bit by opening an issue were we can discuss were to move such things. @julfarn could you open an issue and ping me and @burgholzer there? Please also propose something already in the issue description.

burgholzer:
> > I do not want to make a big discussion out of this, but I think one could discuss about the naming of the pass and where it is currently placed. I thought this is exactly the kind of thing the `aux` dialect is for. Yet, this is not placed anywhere near that and is also called `Tmp` instead of `aux`. Feels slightly inconsistent to the inner monk. But in the end, I do not care that much. If all the other points are addressed, this can be merged from my side.
> 
> 
> 
> Actually I have a similar feeling. On the other hand I want to merge it ASAP. We can probably defer it a little bit by opening an issue were we can discuss were to move such things. @julfarn could you open an issue and ping me and @burgholzer there? Please also propose something already in the issue description.

That is fine with me. I do not want to be a blocker here.

mlir/lib/Conversion/AffineRaise/SCFToAffine.cpp
julfarn: added the check

mlir/lib/Conversion/AffineRaise/WhileToFor.cpp
julfarn: added fail check

julfarn:
I opened issue #109 for the moment (very barebones still, should be shaped up a bit).

burgholzer (approved):
LGTM

PR #93, opened 2026-06-22
Emit QIR

### LINKED ISSUE #87: Include translation from MLIR to LLVM in QCC
For pipelines emitting QIR, `qcc` should ouput actual QIR. Right now one has to run 
```
qcc <input> | mlir-translate --mlir-to-llvmir -o <output>
```
This should be included in `qcc`. (perhaps via an option `--emit-qir`).

rainij:
Resolves #87:

- Add options --target, --compile-to, and --binary.
- Right now the only supported target is QIR. But we set up already the scaffolding.

I took inspiration from existing tools like llc, gcc, clang, iree. The default is to compile to LLVM IR, which implies QIR with the default target.


mlir/test/tools/qcc/qrisp-integration/README.md
```diff
@@ -36,10 +36,11 @@ This guide documents the process of converting a [Qrisp](https://qrisp.eu/) quan
 
 ## Notes on checks
 
-- You may want to check both `QIR` and `qirrunner` simulation output:
+- You may want to check both `QIR` and `qirrunner` simulation output. `qcc` emits
+  QIR (LLVM IR) directly by default, so no `mlir-translate` step is needed:
```
julfarn: I don't think we need this explanation why we don't need something (anymore). This is only relevant info for devs who are expecting `qcc` to emit MLIR, which won't exist in a short time.

mlir/tools/qcc/qcc.cpp
```diff
@@ -29,16 +30,35 @@
...
 
 static cl::OptionCategory qccCategory("QCC options");
 
+namespace {
+/// The target determines the backend to compile for, the actual passes
+/// (pipeline), and the runtime.
+enum class Target : uint8_t { Qir, HisepQ };
+
+/// The stage to compile to and emit.
+enum class Stage : uint8_t { Mlir, LlvmIr, Native };
```
julfarn, line 59: Is there good precedent for the term "stage"? I assume this refers to the level of abstraction of the output. (I always like the term "level", because it's the L in MLIR.) The meaning is not immediately obvious from the name. 

As for `Mlir` next to `LlvmIr`, it feels a little bit awkward. There's no lowering going on, it feels more similar to the difference between textual and bytecode output. Not blocking though, as I don't think there's a perferct solution here.

mlir/tools/qcc/qcc.cpp
```diff
@@ -92,20 +139,58 @@ int main(int argc, char** argv) {
   if (mlir::failed(mlir::applyPassManagerCLOptions(pm))) {
     return 1;
   }
+
+  // Only the QIR pipeline exists today; --target was already validated above
+  // (HiSEP-Q is rejected), so there is nothing to switch on here yet.
```
julfarn: Rather than documenting what's the current state, I would explain what needs to be done in the future (also includes a TODO tag):


```suggestion
  // TODO: Assemble pipeline based on target once a second valid target is added.
```

mlir/tools/qcc/qcc.cpp
```diff
@@ -92,20 +139,58 @@ int main(int argc, char** argv) {
...
     llvm::errs() << errorMessage << "\n";
     return 1;
   }
 
-  module->print(output->os());
-  output->keep(); // otherwise file gets deleted
+  // Refuse to dump binary onto a terminal:
+  if (binary && llvm::CheckBitcodeOutputToConsole(outFile->os())) {
+    return 1;
+  }
```
julfarn, line 159: This needs a helpful error message. Something like this:


```suggestion
  if (binary && llvm::CheckBitcodeOutputToConsole(outFile->os())) {
    llvm::errs() << "cannot display binary output\n";
    llvm::errs() << "Hint: The compilation was successful. Recompile without the --binary flag to print textual form to screen, or specify an output file with the -o option.\n";
    return 1;
  }
```

mlir/tools/qcc/qcc.cpp
```diff
@@ -92,20 +139,58 @@ int main(int argc, char** argv) {
...
+
+  switch (compileTo) {
+  case Stage::Mlir:
+    if (binary) {
+      if (mlir::failed(mlir::writeBytecodeToFile(*module, outFile->os()))) {
+        llvm::errs() << "failed to write MLIR bytecode\n";
```
julfarn, line 165:
```suggestion
        llvm::errs() << "failed to write generated MLIR bytecode to file\n";
```

Trying to eliminate any ambiguity from this error message. I want to know immediately that the bytecode was successfully generated, it's just that the file can't be written.

mlir/tools/qcc/qcc.cpp
```diff
@@ -92,20 +139,58 @@ int main(int argc, char** argv) {
...
+      llvm::WriteBitcodeToFile(*llvmModule, outFile->os());
+    } else {
+      llvmModule->print(outFile->os(), /*AAW=*/nullptr);
+    }
+    break;
+  }
+  case Stage::Native:
+    llvm_unreachable("the 'native' stage is rejected during option validation");
+  default:
+    llvm_unreachable("--compile-to should always have a value (default value if nothing is set explicitly)");
```
julfarn, line 189:
```suggestion
    llvm_unreachable("--compile-to will always have a value (default value if nothing is set explicitly)");
```

When using `llvm_unreachable` we should be highly confident and treat the message as a comment for justification.

julfarn (changes requested):
I only have nitpicks. 

The general approach of having two axes for compilation arguments (how low? and which pipeline?) makes sense. 

My main complaints are the name "Stage" (though I now realize it's not user-facing and therefore not so relevant) and a missing error message. 

When this gets more complicated, we should move things out of the main function, but this will naturally happen.

mlir/test/tools/qcc/qrisp-integration/README.md
rainij: Resolved by resolving conflicts with main which already had an updated version of the README.

mlir/tools/qcc/qcc.cpp
EnriqueNB, line 59: It is also not clear to me what native stands for. Is it the same as the selected target?

rainij: Stage vaguely comes from lingo surrounding compilers like gcc. IMO Mlir next to LlvmIr is not awkward. Basically we differentiate the boundaries between the 3 large stages:

- frontend plus optimization in MLIR.
- backend stuff inside llvm world.
- final output as QISA.

these are three different worlds with different tool sets. I guess that the first two will blur in the future, but there is still a large gap.

But indeed we could rename it: Maybe `CompileTo`?

mlir/tools/qcc/qcc.cpp
rainij: No this already prints something:

```
$ qcc --binary mlir/test/tools/qcc/cli_options.mlir 
WARNING: You're attempting to print out a bitcode file.
This is inadvisable as it may cause display problems. If
you REALLY want to taste LLVM bitcode first-hand, you
can force output with the `-f' option.
```

so I leave it as is.

mlir/tools/qcc/qcc.cpp
rainij: I actually prefer the shorter one. I am not 100% sure what exactly `mlir::writeBytecodeToFile(*module, outFile->os())` does. We only know that it takes a module and an output stream. I do not really know if the bytecode generation could fail here and I do not want to imply any knowledge (I believe that the in-memory representation of a module is _not_ bytecode). I would also leave out the "to file". I use the wording in the code because it is often understood that edge cases were it isn't an actual file are possible. I am just naming the common case. But on the commandline the user knows when they just wanted to print to screen, then talking about a "file" seems wrong.

mlir/tools/qcc/qcc.cpp
rainij: I have to strongly disagree, "should" is really the right word here. There are multiple reasons:

About correctness: We state here that any code which reaches this point violates our spec. We should not and cannot guarantee that this line can never be reached by a program (I definitely would read "will" as this claim). It could have a bug. Otherwise you could ask: why including the line in the first place? Why not just adding a comment?

Consistency: In LLVM itself "should" is by far the more used idiom then "will". Changing it to "will" might even invite more "bikeshed" discussions. I also know it outside of llvm, and also for testing, not just assertions.

I asked claude to check my claim:

```
Done — whole repo: 9,075 llvm_unreachable messages.

  ┌──────────────────────────┬─────┐
  │         phrasing         │  n  │
  ├──────────────────────────┼─────┤
  │ should (positive)        │ 268 │
  ├──────────────────────────┼─────┤
  │ shouldn't / should not   │ 138 │
  ├──────────────────────────┼─────┤
  │ must                     │ 93  │
  ├──────────────────────────┼─────┤
  │ will                     │ 9   │
  ├──────────────────────────┼─────┤
  │ cannot / can't / can not │ 229 │
  ├──────────────────────────┼─────┤
  │ expected                 │ 155 │
  └──────────────────────────┴─────┘

  Exact construction:
  - should always: 4
  - will always: 0
  - must always: 3
```

I even asked it about `cannot` and `must` which also feel strong, but checked that only 11 of the "cannot" are about impossibility of something. They are more in the spirit of missing capability: "Cannot convert type to floating-point". And "must" is about preconditions. Here is a table (by claude) I mostly agree with:

```
  ┌───────────────────┬───────────────────────┬───────────────────────────────────────────┬───────────────────────────────────────────────────────────────┐
  │       verb        │       modality        │                   role                    │          competes with "should always have a value"?          │
  ├───────────────────┼───────────────────────┼───────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
  │ should            │ normative, soft       │ "the contract expects this"               │ yes — same axis                                               │
  ├───────────────────┼───────────────────────┼───────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
  │ must              │ normative, mandatory  │ "the contract requires this"              │ yes — same axis, one notch up                                 │
  ├───────────────────┼───────────────────────┼───────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
  │ expected          │ normative             │ "this was expected to hold"               │ yes                                                           │
  ├───────────────────┼───────────────────────┼───────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
  │ will              │ predictive, factual   │ "this is what happens"                    │ off-axis (overclaims)                                         │
  ├───────────────────┼───────────────────────┼───────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
  │ cannot (218)      │ capability            │ "this operation is impossible"            │ no — different construct                                      │
  ├───────────────────┼───────────────────────┼───────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
  │ cannot/can't (11) │ factual impossibility │ "this state can't occur" — the bare idiom │ will-spirited, but a fixed idiom, not "X cannot have a value" │
  └───────────────────┴───────────────────────┴───────────────────────────────────────────┴───────────────────────────────────────────────────────────────┘
```

mlir/tools/qcc/qcc.cpp
rainij: No it is not the same. If the target is `hisep-q` then native would be its assembly format or an object file (if the `--binary` flag is set).

But these flags are not set in stone. If we later feel they aren't quite right we can still change things.

julfarn: No, I think `CompileTo` would be worse. I have now gotten used to `Stage` and have no better idea. 

I'm for leaving as is, resolve when read.

mlir/tools/qcc/qcc.cpp
julfarn: Huh, okay. I should have checked. This is actually unfortunate because it's incorrect. QCC doesn't support the -f option at the moment. 

Agree with leaving as is.

mlir/tools/qcc/qcc.cpp
julfarn: I misread. I overlooked that the function is also responsible to generate the bytecode. 

Let's leave this as is as well.

mlir/tools/qcc/qcc.cpp
rainij: Indeed surprising. I think the function should not take the responsibility to emit the message, but we get what we get.

mlir/tools/qcc/qcc.cpp
julfarn: I will submit to this research of couse :D  

Just to explain my reasoning: 

> Otherwise you could ask: why including the line in the first place? Why not just adding a comment?

`llvm_unreachable` is used for optimization. In release builds, the compiler treats this branch as literally unreachable, and will perform optimizations under that assumption. Contrasting this with

> We should not and cannot guarantee that this line can never be reached by a program (I definitely would read "will" as this claim).

If we're not ready to make that guarantee, one could argue we shouldn't use `llvm_unreachable` ever. 

But of course it makes sense to stick with the convention.

julfarn (approved):

burgholzer (approved):
LGTM 👍🏼

PR #94, opened 2026-06-23
Relax license header checks for test folder

### LINKED ISSUE #89: Relax license header requirements in python files
Python files that are tests (e.g. qrisp programs) are currently required to have license headers. 

We should look into relaxing this requirement. If it doesn't make sense for all python files, then perhaps we could exclude a particular naming pattern. 

@denialhaag, do you have some opinions on this?

If we decide to drop the headers, do not forget to look over the test folder and remove were the were already added due to CI constraints.

Conclusion of the discussion below summarized in this comment: https://github.com/FullStaQD/compiler/issues/89#issuecomment-4756608859

rainij:
Resolves #89.

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -20,6 +20,9 @@ header:
     - "**/*.py"
     - "**/*.td"
 
+  paths-ignore:
+    - "mlir/test/**/*.py"
+
   comment: on-failure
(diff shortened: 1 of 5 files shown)


burgholzer (approved):
LGTM

PR #97, opened 2026-06-24
Centralize install location of qir-runner

### LINKED ISSUE #92: Install qir-runner on github actions runner
[qir-runner](https://github.com/qir-alliance/qir-runner) should be available for integration tests.

Resolution to the discussion here: https://github.com/FullStaQD/compiler/issues/92#issuecomment-4759358290

rainij:
Resolves #92 .

Lit config checks if `qir-runner` is installed. If not installs it via `uvx` (or fails if that one is not available).


mlir/test/lit.cfg.py
```diff
@@ -79,5 +58,14 @@ class ConfigType:
...
-    raise RuntimeError(msg)
+    lit_config.fatal(f"Could not find qcc and qcc-opt anywhere under {base_tool_dir}.")
+
+# If `qir-runner` is not already available in the environment, fall back to
+# running it ephemerally via `uvx`.
+if shutil.which("qir-runner", path=config.environment["PATH"]) is None:
+    if shutil.which("uvx", path=config.environment["PATH"]) is None:
+        lit_config.fatal(
+            "Could not find the 'qir-runner' executable, which is required to run some tests. "
+            "Either install it yourself, or install 'uvx' (see README) to run it ephemerally instead."
```
julfarn: Is there a way to install `uvx` without installing `uv`? Is there a way to install `uv` without installing `uvx`? We require `uv`, and that should be mentioned here. To my knowledge, `uvx` is just an abbreviation for `uv tool run`. If this could ever cause troubles, we should revert to explicitly use `uv tool run`.

README.md
```diff
@@ -43,6 +43,8 @@ IDEs and the cli should provide you with a way to choose which config you want t
 
 If you prefer to build outside the container, you need LLVM/MLIR version >= [VERSION_USED_BY_DEVCONTAINER_IMAGE](.devcontainer/Dockerfile) installed. If you have no previous installation, check the guide [here](https://mqt.readthedocs.io/projects/core/en/latest/installation.html#setting-up-mlir).
 
+We also recommend to [install](https://docs.astral.sh/uv/getting-started/installation/) `uv` and `uvx`.
```
julfarn: Would just put `uv` here, see other comment.

mlir/test/lit.cfg.py
```diff
@@ -1,32 +1,11 @@
...
-        project_tools_dir: Path = ""
-        llvm_tools_dir: str = ""
-        llvm_shlib_ext: str = ""
-        cmake_build_type: str = ""
-        environment: dict[str, str] = {}
-        substitutions: list[tuple[str, str]] = []
-        excludes: list[str] = []
-
-
-    config = ConfigType()
```
julfarn, line 29: Personally I hate the nature of these config scripts where you just have to know what options are available and what their types are. But I see that having this stub is also distracting and goes against the case of using python here at all. 

On the other hand if you have python tooling installed the amount of type checker warnings is a bit annoying if you remove this.

mlir/test/lit.cfg.py
burgholzer: `uvx` directly comes with `uv`. There is no way to "install uvx (without uv)".
In my opinion, we should be looking for `uv` here and simply doing `uv tool install qirrunner` without the substitution. This makes the `qir-runner` binary available.
(or search for `uv` and execute `uv tool run ...` as a substitution)

README.md
burgholzer: Agree. It's a single tool to be installed.

mlir/test/lit.cfg.py
rainij: Removed mentioning of `uvx`, search for `uv` only and substitute `uv tool run ...` instead. I like the substitution approach a bit more, seems less intrusive.

README.md
rainij: Done.

mlir/test/lit.cfg.py
rainij: These things always feel like a lie to me when it is not backed by proper tooling. I already stumble over `project_tools_dir: Path` and then assigning it to a string anyway. My feeling was that this was added to "shut up the type checker". That is not a good reason to add it. The previous code also hardcoded pyright, we never decided on this.

The only minor benefit is that you get code completion. But then I ask, how much time does it save in real live (in our particular scenario)? And compare that to, how much time do I spend maintaining this thing and discussing minor errors in it.

To really understand what is going on you have to look into the sources anyway `/llvm-project/llvm/utils/lit/` or the docs at https://llvm.org/docs/CommandGuide/lit.html#test-suites . I know it is not super duper SOTA way to do it, but this is something which can only properly be fixed upstream and I guess nobody feels the urge.

burgholzer (approved):
LGTM 👍🏻

julfarn (approved):

PR #98, opened 2026-06-24
QIR -> LLVM Intrinsics -> HiSEPQ RISCV Assembly.

### LINKED ISSUE #49: Pass for HiSEP-Q Intrinsics
In general the intrinsics look like this

```llvm
; QV.SINGLE: qv.h vs1, rs2, Block_imm
declare void @llvm.riscv.qv.h.nxv8i8.i32.i32.i32(<vscale x 8 x i8>, i32, i32, i32)
; QV.PAIR: qv.cx vs1, vs2, Block_imm
declare void @llvm.riscv.qv.cx.nxv8i8.nxv8i8.i32.i32(<vscale x 8 x i8>, <vscale x 8 x i8>, i32, i32)
```

Note that the suffix `.nxv8i8.i32.i32.i32` comes from mangling in the overloaded argument types. In RISC-V vector extension lingo we have that $\mathrm{vscale} = \mathrm{VLEN} / 64$ (e.g. $2$ for `VLEN=128`). Moreover, for a type `<vscale x A x iB>` we have

$$
  \mathrm{LMUL} = \frac{A \times B}{64} = \frac{\mathrm{vscale} \times A \times B}{\mathrm{VLEN}} .
$$

In the first iteration we just want to make the conversion work. For this reason 

- Start from the output of the QIR lowering (already in LLVM dialect).
- Replace each gate by an intrinsic where the vector has just one element.

_Later_ we might want to skip the QIR step entirely (to be discussed) and we _must_ devise a plan how to actually utilize the SIMD nature of the ISA (either via explicit frontend control, an optimization pass, ..., or all of that). But this is not part of this issue.

### LINKED ISSUE #68: Erase record statements right before codegen
This is a workaround.

In the LLVM backend, during HiSEP-Q codegen, whenever you see a record statement like `aux.record_bool`, ignore it. Put a TODO comment somewhere into the code making clear this is a workaround.

Background: A proper implementation should come later. The HiSEP-Q team has to do some research first. E.g. looking into the xilinx runtime library.

EnriqueNB:
Conversion passes to go from QC/QIR to LLVM HiSEPQ Intrinsics have been implemented. This allows to lower from a .mlir file to an assembly file .s using the pipeline:

```
qcc-opt | mlir-translate | llc
```

Address #49.


mlir/include/qcc/Conversion/ToIntrinsics/ToIntrinsics.td
rainij, line 1: Why are there two conversions? QIR to intrinsics, QC to intrinsics. We should keep the PR focused to QIR to intrinsics. Bypassing QIR is still an option but this should be a different PR. Also, QC might not be the right starting point. If this is the case the other pass would also be blocked from being merged to main.

mlir/include/qcc/Conversion/Aux_/AuxOutputRecording.td
rainij, line 1: Why is this file reformatted? Also note that `aux.record_bool` was recently removed in favor of a more general `aux.record_int`.

rainij (changes requested):
Please always fix the CI first. There failing tests and also linter errors.

mlir/include/qcc/Conversion/ToIntrinsics/ToIntrinsics.td
EnriqueNB: I added the QIR to intrinsics because I started the merge request by copying the QCtoQIR pass. Since we talked about having also the direct route I kept this route but I agree that this should be addressed in another MR.

EnriqueNB:
The MR is ready for review. I added the options to `qcc` to compile to `assembly` and `object` code: 

`qcc --target=hisep-q --compile-to=assembly ghz.mlir -o ghz.s` 
`qcc --target=hisep-q --compile-to=object ghz.mlir -o ghz.o`

The CI is failing because it doesn't have RISCV as supported LLVM target.

rainij (changes requested):
I am OOO this week, but you can also ask Julian for a first review. But before that check the CI to make sure the tests pass.

EnriqueNB:
The branch has been rebased and now it pass the CI. @rainij please review.

mlir/include/qcc/Conversion/ToIntrinsics/ToIntrinsics.td
```diff
@@ -0,0 +1,45 @@
...
+      carrying that index in lane 0, matching the `anyvector` operand that the
+      QVSingle / QVPair intrinsics expect.
+
+      Additionally:
+      - `__quantum__rt__initialize` calls are erased (no-op on the intrinsic path).
+      - `__quantum__rt__read_result` calls are replaced with `undef : i1`
+        (TODO: a proper `qv.read_result` intrinsic is not yet available).
+      - `__quantum__rt__bool_record_output` and `__quantum__rt__int_record_output`
+        calls are erased.
+      - Leftover QIR function declarations are removed from the module.
```
rainij: I would move this section to the bottom and maybe replace "Additionally" by "Additional simplifying assumptions". Then I would move the TODO and phrase it more generally. I think in general that these concepts have no counterpart on the ISA side shows that it's limited expressiveness. Currently HiSEP-Q focusses on the application side of things (user program) but cannot express any maintenance operations (initialization is kind of maintenance). But even on the application side there are blind spots like output recording. Maybe you can find a good formulation for this general issue (as a TODO comment).

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,240 @@
...
+      return failure();
+    }
+
+    rewriter.replaceOpWithNewOp<LLVM::UndefOp>(callOp, rewriter.getI1Type());
+    return success();
+  }
+};
+
+/// Erases `llvm.call @__quantum__rt__initialize(ptr)`.
+/// The runtime initialization step is not needed on the bare-metal intrinsic path.
```
rainij: This isn't exactly true. The QIR terminology "rt init" might be misleading though. Here is what the [spec says](https://github.com/qir-alliance/qir-spec/blob/main/specification/profiles/Adaptive_Profile.md#runtime-functions):

```
Initializes the execution environment. Sets all qubits to a zero-state if they are not dynamically managed.
```

So qubit initialization gets removed by this.

So a TODO stating that we erase something our QISA cannot represent would be more appropriate. IMO should be also added to the other patterns which just erase stuff.

mlir/lib/Target/HiSEPQ/CMakeLists.txt
```diff
@@ -1,4 +1,19 @@
...
+    RISCVInfo
+    CodeGen
+    MC
+    Target
+    SelectionDAG
+    AsmPrinter)
+
+add_mlir_library(
+  QCCTargetHiSEPQ HiSEPQTarget.cpp LINK_LIBS MLIRPass
+  # Shared QIR lowering + the QIR->intrinsics conversion:
```
rainij: Do we need this comment? Also, we usually format this stuff a bit differently.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
```diff
@@ -0,0 +1,127 @@
+// ===----------------------------------------------------------------------===//
+//
+// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM
+// Exceptions.
+// See <repo-root>/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// ===----------------------------------------------------------------------===//
+
```
rainij:
```suggestion
```

Never add license headers to mlir test files.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
```diff
@@ -0,0 +1,127 @@
...
+// CHECK-DAG:     %[[ZERO:.*]] = llvm.mlir.constant(0 : i32) : i32
+// CHECK-DAG:     %[[ONE:.*]] = llvm.mlir.constant(1 : i32) : i32
+// CHECK-DAG:     %[[UNDEF_VEC:.*]] = llvm.mlir.undef : vector<[8]xi8>
+// CHECK:         %[[VEC0:.*]] = llvm.insertelement %[[IDX0]], %[[UNDEF_VEC]][%[[ZERO]] : i32] : vector<[8]xi8>
+// CHECK:         llvm.call_intrinsic "llvm.riscv.qv.h"(%[[VEC0]], %[[ZERO]], %[[ZERO]], %[[ONE]])
+// CHECK:         %[[VEC1:.*]] = llvm.insertelement %[[IDX1]], %[[UNDEF_VEC]][%[[ZERO]] : i32] : vector<[8]xi8>
+// CHECK:         llvm.call_intrinsic "llvm.riscv.qv.x"(%[[VEC1]], %[[ZERO]], %[[ZERO]], %[[ONE]])
+
+
+llvm.func @cx_gate() attributes { passthrough = ["entry_point"] } {
```
rainij: Rename to `@two_qubit_gates`?

mlir/test/tools/qcc/example_ghz_intrinsics.mlir
```diff
@@ -0,0 +1,65 @@
+// ===----------------------------------------------------------------------===//
+//
+// Part of the FullStaQD Project, under the Apache License v2.0 with LLVM
+// Exceptions.
+// See <repo-root>/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// ===----------------------------------------------------------------------===//
+
```
rainij:
```suggestion
```

No license headers in mlir files.

mlir/test/tools/qcc/example_ghz_intrinsics.mlir
```diff
@@ -0,0 +1,65 @@
...
+// REQUIRES: hisep-q
+
+// RUN: qcc --target=hisep-q --compile-to=mlir -o - %s | FileCheck %s
```
rainij: This test could be moved into the file `exmaple_ghz.mlir`. Add this RUN line over there and use a dedicated CHECK prefix.

mlir/tools/qcc/qcc.cpp
```diff
@@ -197,8 +203,22 @@ int main(int argc, char** argv) {
     }
     break;
   }
-  case Stage::Native:
-    llvm_unreachable("the 'native' stage is rejected during option validation");
+  case Stage::Native: {
+    // Native emission is delegated to the target backend (the only place allowed
+    // to depend on target-specific/fork codegen). `--binary` selects an object
+    // file; otherwise textual assembly is emitted.
```
rainij:
```suggestion
```

This comment is not useful IMO.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -11,16 +11,86 @@
...
+
+bool emitNativeHiSEPQ(llvm::Module& module, llvm::raw_pwrite_stream& os, const NativeCodegenOptions& options) {
+  // HiSEP-Q QISA is encoded as the experimental "xqv" RISC-V vector extension,
+  // provided by the HiSEP-Q LLVM fork.
+  LLVMInitializeRISCVTargetInfo();
+  LLVMInitializeRISCVTarget();
+  LLVMInitializeRISCVTargetMC();
+  LLVMInitializeRISCVAsmPrinter();
+
+  const std::string attrsStr = options.mattr.empty() ? "+experimental-xqv" : options.mattr.str();
```
rainij: Making it possible to override `+experimental-xqv` seems really wrong here. Please remove the option here.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -11,16 +11,86 @@
...
+  llvm::Triple triple(
+      llvm::Triple::normalize(options.mtriple.empty() ? "riscv32-unknown-unknown" : options.mtriple.str()));
```
rainij: Similarly as above, please remove the option here. It seems wrong.

mlir/tools/qcc/qcc.cpp
```diff
@@ -80,10 +80,16 @@ int main(int argc, char** argv) {
...
+                 clEnumValN(Stage::Native, "native", "Native target code (QISA; requires a target with a backend)")),
       cl::cat(qccCategory));
   const cl::opt<bool> binary("binary", cl::desc("Emit the binary encoding (obj/bytecode/bitcode) instead of text"),
                              cl::init(false), cl::cat(qccCategory));
+  const cl::opt<std::string> mtriple("mtriple",
+                                     cl::desc("Target triple for native code generation (target default if empty)"),
+                                     cl::init(""), cl::value_desc("triple"), cl::cat(qccCategory));
+  const cl::opt<std::string> mattr(
+      "mattr", cl::desc("Target attribute string for native code generation (target default if empty)"), cl::init(""),
+      cl::value_desc("attrs"), cl::cat(qccCategory));
```
rainij: Nothing tests "mtriple" and "mattr". In the spirit of YAGNI please remove. The current PR really doesn't need it.

mlir/include/qcc/Target/TargetRegistry.h
```diff
@@ -16,7 +16,22 @@
...
 
+namespace llvm {
+class Module;
+class raw_pwrite_stream;
+} // namespace llvm
+
 namespace qcc {
+
+/// Options controlling native (QISA) code emission for a target.
+struct NativeCodegenOptions {
```
rainij, line 27: Please remove `mtriple` and `mattr`. See also my other comments on those two.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,240 @@
...
+    auto loc = callOp.getLoc();
+    auto i32Type = rewriter.getI32Type();
+    auto operands = callOp.getArgOperands();
+
+    Value blockImm = LLVM::ConstantOp::create(rewriter, loc, i32Type, rewriter.getI32IntegerAttr(0));
+    Value vl = LLVM::ConstantOp::create(rewriter, loc, i32Type, rewriter.getI32IntegerAttr(1));
+
+    SmallVector<Value> args;
+
+    if (*callee == qcc::qirQisCX) {
```
rainij: This if-else basically hardcodes "CX is the only 2-qubit gate". This is brittle. Please refactor this into something which survives adding another 2-qubit gate.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,240 @@
...
+/// Encodes a qubit index as a `vector<[8]xi8>` scalable vector for QV intrinsics.
+/// The index is inserted into lane 0 of an undef vector. The `nxv8i8` element
+/// type matches the HiSEP-Q RISC-V backend's QV instruction-selection patterns.
+static Value qubitIndexToVec(OpBuilder& builder, Location loc, int64_t index) {
+  auto i8Type = builder.getIntegerType(8);
+  auto i32Type = builder.getI32Type();
+  auto vecType = VectorType::get({8}, i8Type, /*scalableDims=*/{true});
+
+  Value indexConst = LLVM::ConstantOp::create(builder, loc, i8Type, builder.getIntegerAttr(i8Type, index));
+  Value undef = LLVM::UndefOp::create(builder, loc, vecType);
```
rainij: It seems `undef` is soft deprecated. Please use `poison` instead. I mention it just here, but please refactor all uses of undef.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+/// Encodes a qubit index as a `vector<[8]xi8>` scalable vector for QV intrinsics.
+/// The index is inserted into lane 0 of an undef vector. The `nxv8i8` element
+/// type matches the HiSEP-Q RISC-V backend's QV instruction-selection patterns.
+static Value qubitIndexToVec(OpBuilder& builder, Location loc, int64_t index) {
+  auto i8Type = builder.getIntegerType(8);
+  auto i32Type = builder.getI32Type();
+  auto vecType = VectorType::get({8}, i8Type, /*scalableDims=*/{true});
+
+  Value indexConst = LLVM::ConstantOp::create(builder, loc, i8Type, builder.getIntegerAttr(i8Type, index));
```
rainij, line 104: The `index` which is 64 bits is converted to 8 bits here. The compiler should emit an error on integer wrap. It does not necessarily need to happen here, but this function could have an `assert` at the very least (in this case something else needs to emit the error).

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,240 @@
...
+    if (intrName.empty()) {
+      return failure();
+    }
+
+    auto loc = callOp.getLoc();
+    auto i32Type = rewriter.getI32Type();
+    auto operands = callOp.getArgOperands();
+
+    Value blockImm = LLVM::ConstantOp::create(rewriter, loc, i32Type, rewriter.getI32IntegerAttr(0));
+    Value vl = LLVM::ConstantOp::create(rewriter, loc, i32Type, rewriter.getI32IntegerAttr(1));
```
rainij, line 164: These two lines already mutate the IR, this is not allowed at this point as you have not yet decided whether the pattern matches. Down below you have have this if statement which still allows to emit failure (emitError).

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
```diff
@@ -0,0 +1,127 @@
...
+// CHECK-DAG:     %[[TIDX:.*]] = llvm.mlir.constant(3 : i8) : i8
+// CHECK-DAG:     %[[ZERO:.*]] = llvm.mlir.constant(0 : i32) : i32
+// CHECK-DAG:     %[[ONE:.*]] = llvm.mlir.constant(1 : i32) : i32
+// CHECK-DAG:     %[[UNDEF_VEC:.*]] = llvm.mlir.undef : vector<[8]xi8>
+// CHECK:         %[[CVEC:.*]] = llvm.insertelement %[[CIDX]], %[[UNDEF_VEC]][%[[ZERO]] : i32] : vector<[8]xi8>
+// CHECK:         %[[TVEC:.*]] = llvm.insertelement %[[TIDX]], %[[UNDEF_VEC]][%[[ZERO]] : i32] : vector<[8]xi8>
+// CHECK:         llvm.call_intrinsic "llvm.riscv.qv.cx"(%[[CVEC]], %[[TVEC]], %[[ZERO]], %[[ONE]])
+
+
+llvm.func @measurement() -> i1 attributes { passthrough = ["entry_point"] } {
```
rainij, line 71: Please add a TODO comment that HiSEP-Q does not yet provide a way to record anything. This makes it clear that this test can safely be modified in the future once a better solution arrives.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
```diff
@@ -0,0 +1,127 @@
...
+
+// RUN: qcc-opt %s -convert-qir-to-intrinsics | FileCheck %s
+
+// Input: a module as produced by the ToQIR pipeline.
+// Each qubit is an `!llvm.ptr` obtained via `llvm.inttoptr` of a constant index.
+// QIS gate calls use those ptrs as operands.
+//
+// Expected output: every QIS call is replaced by an `llvm.call_intrinsic` call
+// with the qubit encoded as a `vector<[8]xi8>` scalable vector in lane 0.
+// Constants and undefs are CSE'd into a shared preamble within each function.
```
rainij: This last line shares a technical detail of the pass driver I believe. Should probably not be mentioned here.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,240 @@
...
+/// type matches the HiSEP-Q RISC-V backend's QV instruction-selection patterns.
+static Value qubitIndexToVec(OpBuilder& builder, Location loc, int64_t index) {
+  auto i8Type = builder.getIntegerType(8);
+  auto i32Type = builder.getI32Type();
+  auto vecType = VectorType::get({8}, i8Type, /*scalableDims=*/{true});
+
+  Value indexConst = LLVM::ConstantOp::create(builder, loc, i8Type, builder.getIntegerAttr(i8Type, index));
+  Value undef = LLVM::UndefOp::create(builder, loc, vecType);
+  Value lane = LLVM::ConstantOp::create(builder, loc, i32Type, builder.getI32IntegerAttr(0));
+  return {LLVM::InsertElementOp::create(builder, loc, undef, indexConst, lane)};
```
rainij:
```suggestion
  Value insertElmOp = LLVM::InsertElementOp::create(builder, loc, undef, indexConst, lane);
  return insertElmOp;
```

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,240 @@
...
+    Value vl = LLVM::ConstantOp::create(rewriter, loc, i32Type, rewriter.getI32IntegerAttr(1));
+
+    SmallVector<Value> args;
+
+    if (*callee == qcc::qirQisCX) {
+      // QVPairIntrinsic: (vs1: vec<[8]xi8>, vs2: vec<[8]xi8>, block_imm: i32, vl: i32)
+      auto ctrlIdx = getQubitIndexFromPtr(operands[0]);
+      auto tgtIdx = getQubitIndexFromPtr(operands[1]);
+      if (!ctrlIdx || !tgtIdx) {
+        return callOp.emitError("convert-qir-to-intrinsics: cannot extract qubit index from ptr "
```
rainij: This should actually be a hard error here. As it is it just does not lead to a pass failure, merely to a match failure.

I asked Claude for the standard pattern on how to propagate the error to the pass itself. It claims that one can attach a reference `hadError` to the rewriter which the pass then can readout. Please verify if this is a valid approach (I read somewhere that one should avoid state at a rewriter but claude claims a ref doesn't really count) or find another way to make sure the error is thrown (and tested). Here is the suggestion by Claude

```cpp
  struct QISCallLowering : OpRewritePattern<LLVM::CallOp> {
    QISCallLowering(MLIRContext* ctx, bool& hadError)
        : OpRewritePattern(ctx), hadError(hadError) {}
    ...
      if (!ctrlIdx || !tgtIdx) {
        callOp.emitError("cannot determine static qubit index for '") << *callee << "'";
        hadError = true;
        return failure();
      }
    ...
    bool& hadError;   // lives in the pass, not really pattern state
  };

  // in runOnOperation:
  bool hadError = false;
  patterns.add<QISCallLowering>(ctx, hadError);
  ...
  if (failed(applyPatternsGreedily(moduleOp, std::move(patterns))) || hadError)
    return signalPassFailure();
``

File: mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp [hunk 21]
```diff
@@ -11,16 +11,86 @@
...
+// pulled in (fork-only) via LLVM_LINK_COMPONENTS.
+// NOLINTBEGIN(readability-identifier-naming): these are LLVM C-API entry points
+// whose names are fixed by LLVM and cannot follow the project's naming style.
+extern "C" {
+void LLVMInitializeRISCVTargetInfo();
+void LLVMInitializeRISCVTarget();
+void LLVMInitializeRISCVTargetMC();
+void LLVMInitializeRISCVAsmPrinter();
+}
+// NOLINTEND(readability-identifier-naming)
```
rainij: It took me a while to understand the comment, it seems overly verbose. In my local checkout I replaced this whole thing with 

```cpp
#include "llvm/TargetParser/Triple.h"
```

And I got no linter errors. So is this about the CI pipeline? Does that one fail?

In any case I would like to get rid of these decls. There is a lot of pre-processor magic happening in the include itself, I fear that these decls do not capture everything and are subtly wrong. If even they are correct, they are a complication I would like to avoid.

rainij (changes requested):

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: There is no problem locally but the problem is with the CI pipeline that doesn't recognize the RISCVTarget when  LLVM is built without it. I created this as a workaround.

rainij: Please investigate possible solutions for this. I know the CI now complicates matters but we still need a proper fix.

burgholzer: I'd propose to simply add clang-tidy ignores for the entire directory tree that cannot be tested without custom LLVM in a separate `.clang-tidy` file.
The alternative would be to use the source build of LLVM for the linter job, which would painfully slow down every single PR CI.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
EnriqueNB: Changed.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
EnriqueNB: Line erased.

mlir/include/qcc/Conversion/ToIntrinsics/ToIntrinsics.td
EnriqueNB: Changed.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
EnriqueNB: Added.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
EnriqueNB: Yes, for future extensibility reasons.

mlir/test/tools/qcc/example_ghz_intrinsics.mlir
EnriqueNB: Deleted.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: This is totally true. It was a quite old comment.

mlir/lib/Target/HiSEPQ/CMakeLists.txt
EnriqueNB: Changed.

mlir/tools/qcc/qcc.cpp
EnriqueNB: Deleted.

mlir/test/tools/qcc/example_ghz_intrinsics.mlir
EnriqueNB: Merged.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: Agree because we are inside emitting NativeHiSEPQ. Deleted.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: Deleted.

mlir/tools/qcc/qcc.cpp
EnriqueNB: Agree in the spirit of YAGNI. When needed these lines shall return.

(This comment reads like a fantasy book prophesy :))

EnriqueNB: Deleted.

mlir/include/qcc/Target/TargetRegistry.h
EnriqueNB: Deleted.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Changed.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Changed.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Now this is checked and emits an error.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Refactored.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Implemented as suggested.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Refactored.

EnriqueNB:
@rainij All your comments have been addressed. The only comment left is the one related to the CI but as this is not related to the pass itself I request your review again as the CI change requires some thought.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,295 @@
...
+      args.push_back(vl);
+      break;
+    }
+
+    LLVM::CallIntrinsicOp::create(rewriter, loc, rewriter.getStringAttr(info->intrinsicName), args);
+    rewriter.eraseOp(callOp);
+    return success();
+  }
+
+  bool* hadError; // Points to storage owned by the pass; valid for one applyPatternsGreedily call.
```
rainij:
```suggestion
private:
  bool* hadError;
```

rainij: The comment is not good, it claims something the class itself cannot really guarantee. Nor is it needed.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
rainij: The ignore solution is definitely an option to make it work right now. I just wonder if we should have a follow up which adds linting support for the QISA CI. I am not sure why it should be painfully slow though. We already have to build LLVM for the QISA CI and the CI takes only 6 minutes when LLVM is in cache. But I do remember that clang-tidy is annoying to set up as it is super slow in general (which is why people tend to let it only run on the diff). But maybe in AI world this isn't an issue anymore.

So my proposal is:
- ignore the folder right now and make the code here look nice.
- open a follow up PR to add support for HiSEP-Q linting.

Does this sound like a plan?

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,295 @@
...
+    if (!info) {
+      return failure();
+    }
+
+    auto operands = callOp.getArgOperands();
+    unsigned numQubitOperands = info->getNumQubitOperands();
+
+    if (operands.size() < numQubitOperands) {
+      callOp.emitError("convert-qir-to-intrinsics: '")
+          << *callee << "' expects at least " << numQubitOperands << " qubit operand(s), got " << operands.size();
```
rainij: Please leave out the `convert-qir-to-intrincics:` prefix this is not necessary with the extensive debug framework MLIR provides. The debug output prints the (actual) pass name if you ask it to. This would just add noise and is not LLVM convention. Also, it is likely forgotton to be updated if somebody changes the pass name.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,295 @@
...
+    SmallVector<int64_t> qubitIndices;
+    for (unsigned i = 0; i < numQubitOperands; ++i) {
+      auto idx = getQubitIndexFromPtr(operands[i]);
+      if (!idx) {
+        callOp.emitError("convert-qir-to-intrinsics: cannot extract qubit index from ptr for '") << *callee << "'";
+        *hadError = true;
+        return failure();
+      }
+      if (*idx < 0 || std::cmp_greater(*idx, std::numeric_limits<uint8_t>::max())) {
+        callOp.emitError("convert-qir-to-intrinsics: qubit index ") << *idx << " out of range for '" << *callee << "'";
```
rainij: Please leave out the `convert-qir-to-intrincics:` prefix. See similar comment above.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
```diff
@@ -0,0 +1,295 @@
...
+          << *callee << "' expects at least " << numQubitOperands << " qubit operand(s), got " << operands.size();
+      *hadError = true;
+      return failure();
+    }
+
+    SmallVector<int64_t> qubitIndices;
+    for (unsigned i = 0; i < numQubitOperands; ++i) {
+      auto idx = getQubitIndexFromPtr(operands[i]);
+      if (!idx) {
+        callOp.emitError("convert-qir-to-intrinsics: cannot extract qubit index from ptr for '") << *callee << "'";
```
rainij: Please leave out the `convert-qir-to-intrincics:` prefix. See similar comment above.

mlir/test/tools/qcc/example_ghz.mlir
```diff
@@ -1,9 +1,14 @@
...
+// RUN: %if hisep-q %{ qcc --target=hisep-q --compile-to=mlir -o - %s | FileCheck %s --check-prefix=CHECK-INTRINSICS %}
 
 /// Prepare GHZ state (without loop so far).
 ///
 /// TODO: This must be replaced by a test starting from jasp. Currently the frontend and the
 /// backend do not fit together.
+///
+/// The second RUN line (gated on `hisep-q`) compiles the same circuit with
+/// --target=hisep-q to verify the full pipeline replaces QIS calls with
+/// RISC-V QV intrinsics.
```
rainij:
```suggestion
```

This comment doesn't add any value.

mlir/test/tools/qcc/target-hisep-q-native.mlir
```diff
@@ -0,0 +1,18 @@
+// REQUIRES: hisep-q
+
+// Native QISA emission: `--compile-to=native` emits assembly, `--binary` an object file.
+
```
rainij:
```suggestion
```

This comment does not provide any value.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
```diff
@@ -0,0 +1,164 @@
+// RUN: qcc-opt %s -convert-qir-to-intrinsics --split-input-file --verify-diagnostics | FileCheck %s
+// The invalid cases below must make the pass itself fail, not merely leave the
+// call unconverted, so also assert a non-zero exit status.
```
rainij:
```suggestion
```

This sounds like a justification for a change. This does not belong here.

rainij: Another thing: The `RUN` comments should almost always be the very top thing in the test file. There are exceptions to this rule, but the default should be to not put other comments in between them (except they really fit there).

rainij (changes requested):

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Dropped.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Dropped.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Dropped.

mlir/lib/Conversion/ToIntrinsics/ConvertQIRToIntrinsics.cpp
EnriqueNB: Dropped.

mlir/test/tools/qcc-opt/ToIntrinsics/convert-qir-to-intrinsics.mlir
EnriqueNB: Erased.

mlir/test/tools/qcc/example_ghz.mlir
EnriqueNB: Erased.

mlir/test/tools/qcc/target-hisep-q-native.mlir
EnriqueNB: Erased.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: Sounds good to me. I didn't manage to do the `clang-tidy` approach Lukas suggested. I instead added `cpp-linter-ignore-extra: "mlir/lib/Target/HiSEPQ/**"` to `ci.yml` in the job that was failing.

EnriqueNB:
Everything has been addressed. @burgholzer @rainij please review.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
rainij: Current solution works for me. The only thing: The comment over there in the CI claims that QISA CI does the linting. Is this something you plan to implement?

@burgholzer any general comments on the current solution?

rainij (approved):

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: I thought that `qisa-ci.yml` already ran the Linter. I erased the comment.

burgholzer: Yeah, the CI Changes are fine.
Technically, you should be able to simply copy some of the clang-tidy workflow steps over to the QISA workflow to have that clang-tidy checked as well. In times of AI that should be a simple prompt. 
Can definitely leave that for a follow-up though.

rainij: I made an issue #122 . @EnriqueNB please resolve the thread once you read it.

rainij:
@burgholzer we merged already as @EnriqueNB is in holiday for a long time from tomorrow on. But of course, changes can be made later on if needed.

PR #102, opened 2026-06-25
Add AGENTS.md

### LINKED ISSUE #61: Add AGENTS.md
Create a file `AGENTS.md` (https://agents.md) and add it to version control. This task likely only makes sense for somebody who actually uses an AI agent.

On the other hand gitignore vendor specific files like `CLAUDE.md`. Add a brief hint to the README and mention that each developer can use those vendor specific files for personal customizations.

rainij:
Resolves #61.

Assisted-by: Claude Code


AGENTS.md
```diff
@@ -0,0 +1,29 @@
...
+**qcc** is an MLIR-based compiler that lowers JASP dialect IR (produced by
+[Qrisp](https://qrisp.eu)) to QIR. Generated tools: `qcc-opt` (an
+`mlir-opt`-style pass explorer) and `qcc` (the end-to-end driver).
+
+## Quick commands
+
+Details and alternatives are in [README.md](README.md).
+
+```shell
+# Build (beware that the user might use a different preset)
```
julfarn, line 12: Does it really make sense to have something here that **may** be correct? It't not for me. 

We could also add this file as `AGENTS.md.template` and git-ignore the real thing. This would make more sense to me than having _user-specific_ instructions delegated to _tool-specific_ files.

rainij: > Does it really make sense to have something here that may be correct? It't not for me.

Not sure if I understand. It is definitely correct if you use the the checked in devcontainer. Checking it into vc fits nicely into our "batteries included" approach.

Of course, as soon as you deviate from the standard setup you might want to update your agent settings. As for me I would instruct my claude to read the file only once and generate my own `CLAUDE.md` from it and never look back to `AGENTS.md` again. So it is again an onboarding kind of thing. Similar to the provided devcontainer. You _can_ use it, but you do not have to.

I would not go the `AGENTS.md.template` route. The existing file can already be seen as a template. The only issue I could sense is if somebody has an agent which hardcodes `AGENTS.md` as the agents file. Then they would be in trouble if they want a custom one. I assumed that this edge case is not relevant.

AGENTS.md
```diff
@@ -0,0 +1,29 @@
+# AGENTS.md
+
+Guidance for AI coding agents working in this repository.
+
+## Project
```
burgholzer: Kind of stating the obvious to me. Maybe the following is a bit more concise 
```suggestion
# QCC

```

AGENTS.md
burgholzer: I am also in favor of keeping this as-is.
If someone uses the devcontainer, or simply follows the default setup, this will work out of the box.
If not, then they have made custom changes already and may as well update the agents file.

AGENTS.md
burgholzer, line 1: This is pretty concise, which I like.
The only thing that might be missing is that agents should be instructed to run prek and that all lints should pass.
(If one adds this, then one should also add that this should be run twice on failure because some of the checks have autofixes.)

burgholzer (changes requested):
This is generally looking good. I have a couple of minor comments.

AGENTS.md
rainij: Mention prek now.

burgholzer (approved):
Nice. LGTM

PR #105, opened 2026-06-29
Structured data output recording

ctfle:
Output recording and lowering to QIR for structured data #106 

Implements support for tuples of int64, bool and (flat) memref


mlir/include/qcc/Conversion/ToQIR/Constants.h
```diff
@@ -50,14 +50,32 @@ static constexpr llvm::StringLiteral qirRtResultRecordOutput = "__quantum__rt__r
...
 /// The second parameter defines a string label for the result value. Depending
 /// on the output schema, the label is included in the output or omitted.
 static constexpr llvm::StringLiteral qirRtIntRecordOutput = "__quantum__rt__int_record_output";
 
+/// Adds an array to the generated output.
+///
+/// Signature: `void(i64,ptr)`.
+///
+/// The first parameter defines the size of the array.
+// The second parameter defines a string label for the result value. Depending
```
rainij:
```suggestion
/// The second parameter defines a string label for the result value. Depending
```

mlir/include/qcc/Conversion/ToQIR/Constants.h
```diff
@@ -50,14 +50,32 @@ static constexpr llvm::StringLiteral qirRtResultRecordOutput = "__quantum__rt__r
...
+/// on the output schema, the label is included in the output or omitted.
+static constexpr llvm::StringLiteral qirRtArrayRecordOutput = "__quantum__rt__array_record_output";
+
+/// Adds a tuple to the generated output.
+///
+/// Signature: `void(i64,ptr)`.
+///
+/// The first parameter defines the size of the tuple.
+// The second parameter defines a string label for the result value. Depending
```
rainij:
```suggestion
/// The second parameter defines a string label for the result value. Depending
```

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
```diff
@@ -20,8 +20,10 @@ include "AuxDialect.td"
 
 class AuxOp<string mnemonic, list<Trait> traits = []> : Op<AuxDialect, mnemonic, traits>;
 
+def FlatIntMemRef : MemRefOf<[AnyInteger]>;
```
rainij: The name implies this is "flat". Do we enforce something anywhere?

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -20,6 +20,9 @@
 #include "mlir/Transforms/DialectConversion.h"
 
 #include <llvm/Support/Casting.h>
+#include <llvm/Support/raw_ostream.h>
```
rainij: This include is likely not needed. Please check also the other includes.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -31,7 +30,11 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/TypeSwitch.h"
 
-#include <cstdint>
+#include <llvm/Support/raw_ostream.h>
```
rainij: This is again likely not needed. Please also check the other includes.

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir-with-structured-output-data.mlir
rainij, line 1: Try to put the test into the already existing file `convert-qc-to-qir.mlir` (and adjust the name of the test function accordingly). Multiple entry points in the same file are allowed. And even if not, qcc-opt can split the file.

rainij, line 1: The test is to complex. You only test recording of arrays here. 

- simplify the function (we do not need the quantum stuff).
- remove CHECKs which are unrelated to the recording.
- even for the remaining checks, focus on what matters.

This helps to keep the test much more readable. All the other stuff should be tested elsewhere already.

mlir/test/tools/qcc-opt/ToQIR/memref-output-convert-to-qir.mlir
rainij, line 1: Should also be included in the existing test file.

mlir/test/tools/qcc-opt/ToQIR/memref-output-convert-to-qir.mlir
```diff
@@ -0,0 +1,46 @@
...
+llvm.func @__quantum__rt__array_record_output(i64, !llvm.ptr)
+llvm.func @__quantum__rt__read_result(!llvm.ptr {llvm.readonly}) -> i1
+llvm.func @__quantum__qis__mz__body(!llvm.ptr, !llvm.ptr {llvm.writeonly}) attributes {passthrough = ["irreversible"]}
+llvm.func @__quantum__qis__h__body(!llvm.ptr)
+llvm.func @__quantum__qis__x__body(!llvm.ptr)
+llvm.func @__quantum__qis__cx__body(!llvm.ptr, !llvm.ptr)
+llvm.mlir.global internal constant @".qir_dummy_label"("dummy_label\00") {addr_space = 0 : i32}
+
+
+// CHECK-LABEL:  func.func @test() -> i64 attributes {passthrough = ["entry_point", ["output_labeling_schema", "schema_id"], ["qir_profiles", "adaptive_profile"], ["required_num_qubits", "0"], ["required_num_results", "0"]], qcc.entry_point} {
```
rainij: Remove all the complicated stuff from the CHECKs which does not add value to the test. For example this check line should be better `// CHECK-LABEL: func.func @test`.

I also believe that some CHECK lines can be removed entirely, some can be trimmed.

mlir/test/tools/qcc-opt/ToQIR/memref-output-convert-to-qir.mlir
```diff
@@ -0,0 +1,46 @@
...
+// CHECK:     %[[zero:.*]] = llvm.mlir.zero : !llvm.ptr
+// CHECK:     llvm.call @__quantum__rt__initialize(%[[zero]]) : (!llvm.ptr) -> ()
+// CHECK:     %[[alloc:.*]] = memref.alloc() : memref<3xi64>
+// CHECK:     %[[cast:.*]] = builtin.unrealized_conversion_cast %[[alloc]] : memref<3xi64> to !llvm.struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)>
```
rainij: Is it correct to have an unrealized converstion cast in here?

mlir/test/tools/qcc-opt/output-recording.mlir
rainij, line 1: Please also add a test under the `qrisp-integration` folder. It should be really simple, no gates, simpler than `bell_state.py`, just outputing some numbers.

rainij, line 1: Should a single number not be wrapped into a tuple? I am a bit undecided here, but I would like to make a conscious decision. Right now I do not have the time to google around, but maybe you can do some research.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
rainij, line 1: The corresponding `PrepToQIR.cpp` (which is not edited) should declare `__quantum__rt__array_record_output` and `__quantum__rt__tuple_record_output`. Hints that corresponding test is missing.

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
rainij: We have to make sure that the compiler rejects multi-dim stuff. A test is the best way to fix this.

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
```diff
@@ -206,6 +209,26 @@ struct RecordIntLowering : public OpConversionPattern<aux::RecordIntOp> {
...
+
+  LogicalResult matchAndRewrite(aux::RecordTupleOp op, OpAdaptor adaptor,
+                                ConversionPatternRewriter& rewriter) const override {
+    auto loc = op.getLoc();
+    StringRef labelName = qcc::qirDummyLabelGlobalSymbolName;
+
+    auto addressOf =
+        LLVM::AddressOfOp::create(rewriter, loc, LLVM::LLVMPointerType::get(rewriter.getContext()), labelName);
+
+    Type ty = op.getValue().getType();
```
rainij: Unused?

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
rainij: Just see it today, this should be moved right before `retOp.erase();`. Rationale: in the pass can still signal a failure after this line (e.g. when returning something unsupported). In that case the debug output will look weird: on the one hand the function claims to return nothing, on the other hand we still have a return op with some arguments.

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
```diff
@@ -51,4 +53,78 @@ def RecordIntOp : AuxOp<"record_int"> {
...
+        the QIR function.
+    }];
+
+  let arguments = (ins FlatIntMemRef:$value);
+  let results = (outs);
+
+  let assemblyFormat = "$value attr-dict `:` type($value)";
+}
+
+def RecordTupleOp : AuxOp<"record_tuple"> {
```
rainij: Having both "record tuple" and "record memref" is not really a clean solution. On the one hand it literally lifts a low level detail of QIR to our mid-level dialect: To record a value QIR has to do a two step procedure, first emit the size, then the contents. We should instead just emit the value, which is a container anyway.

So I would erase the record tuple op and defer the tuple creation to the QIR pass.

There is still the problem that dynamic and non-flat shapes can silently pass through and create weird errors. I would add a verfifier level constraint to `RecordMemrefOp` that ensures that the shape is 1D and the size is static. Once we properly support dynamic sizes we can relax this. Please also add a corresponding test. This is related to the `FlatIntMemRef`, which does not check the flattness and his misleading (`IntMemRef` would be correct, but I am unsure if one needs it).

rainij (changes requested):

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
ctfle: Yes there is a check when we convert to QIR (which is where the flatness is important since we don't support strides or offsets when iterating through the elements). Edit: I added a verifier which is the cleaner solution 👍

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
ctfle: Right. Removed 👍

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
ctfle: Removed

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
ctfle: Removed 👍

mlir/lib/Conversion/ToQIR/ConvertQCToQIR.cpp
ctfle: Good catch! Changed

mlir/test/tools/qcc-opt/output-recording.mlir
ctfle: I would argue that primitive values should not be wrapped in a tuple. The main purpose of the tuple is structural framing because the output stream is flat. This is not required for a single primitive. It only creates type miss-matches when lowering and unnecessary runtime overhead.

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
ctfle: 👍

ctfle: changed 👍

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
ctfle: I removed tuple recording from the aux dialect. Now tuple recording gets implemented during the conversion to QIR. Moreover, there is now a verifier that checks for static flat memrefs (and a corresponding test).

mlir/include/qcc/Dialect/Aux_/IR/AuxOps.td
rainij: The verifier is nice.

rainij (changes requested):
I resolved the already resolvable threads. There are a few unaddressed threads left. Maybe you overlooked them (only visible in  "Files changed"). There is also one linter complaint open.

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir-with-structured-output-data.mlir
ctfle: Simplified and moved 👍

mlir/test/tools/qcc-opt/ToQIR/memref-output-convert-to-qir.mlir
ctfle: 👍

mlir/test/tools/qcc-opt/ToQIR/memref-output-convert-to-qir.mlir
ctfle: Ok. I made it a bit simpler (maybe we don't need to check the types every time)

mlir/test/tools/qcc-opt/ToQIR/memref-output-convert-to-qir.mlir
ctfle: Yeah this happens because memref was not fully lowered yet. We can do this also in this pass (its just one line). I added it now then this gets replaced with a bunch of llvm operations.

mlir/test/tools/qcc-opt/output-recording.mlir
ctfle: added

mlir/test/tools/qcc/qrisp-integration/multi_output.py
```diff
@@ -0,0 +1,4 @@
+from qrisp import QuantumVariable, h, cx, measure, q_fori_loop
+
```
rainij:
```suggestion
```

mlir/test/tools/qcc/qrisp-integration/multi_output.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: qcc %s | FileCheck %s
```
rainij: Please imitate the same three RUN lines as in all the other qrisp integration tests. Including the qir-runner call.

mlir/test/tools/qcc/qrisp-integration/multi_output.mlir
```diff
@@ -0,0 +1,19 @@
...
+
+// GENERATED FROM QRISP VERSION 0.9.5 git+https://github.com/eclipse-qrisp/Qrisp.git@b81ea2f979d21cd8d600e79d8b0c7066fe7cbe1b
```
rainij: Did the generator script really create this line?

rainij (changes requested):
There is still some cleanup needed in the multi_output test. The windows CI failed, I triggered a re-run, but if it fails again, please have a look.

mlir/test/tools/qcc/qrisp-integration/multi_output.mlir
ctfle: It generated "// GENERATED FROM QRISP VERSION 0.9.5" I added the link to the repo as in the other test files

mlir/test/tools/qcc/qrisp-integration/multi_output.mlir
ctfle: Added the qir-runner lines

mlir/lib/Conversion/Aux_/AuxOutputRecording.cpp
```diff
@@ -58,25 +68,29 @@ struct AuxOutputRecording final : public impl::AuxOutputRecordingBase<AuxOutputR
...
 
-      funcOp.setType(newType);
-
       auto& body = funcOp.getBody().front();
       auto retOp = cast<func::ReturnOp>(body.getTerminator());
       auto oldReturnOperands = retOp.getOperands();
 
       OpBuilder builder(retOp);
 
+      // auto size = static_cast<int64_t>(oldReturnOperands.size());
```
burgholzer: Commented out code --> remove

```suggestion
```

burgholzer (approved):
Except for some leftover commented-out code and the failing CI, this looks good on a quick glance.

mlir/test/tools/qcc-opt/ToQIR/convert-qc-to-qir-with-structured-output-data.mlir
ctfle: Removed

rainij (approved):

PR #108, opened 2026-06-29
Loop unrolling of nested index casts

### LINKED ISSUE #107: Loop unrolling: Nested index casts not working
Nested for loops of this type fail to raise as of #88: 

```

// BEFORE

  %3 = scf.for %arg0 = %c0_i64 to %c4_i64 step %c1_i64 iter_args(%arg1 = %c0_i64) -> (i64)  : i64 {
    ...
    scf.for %arg2 = %c0_i64 to %arg0 step %c1_i64  : i64 {
      ...
    }
    ...
    %51 = arith.ori %arg1, %50 : i64
    scf.yield %51 : i64
  }


// AFTER

%3 = affine.for %arg0 = 0 to 4 iter_args(%arg1 = %c0_i64) -> (i64) {
    %9 = affine.apply affine_map<(d0, d1)[s0] -> (d0 + d1 * s0)>(%c0, %arg0)[%c1]
    %10 = arith.index_cast %9 : index to i64
    ...
    scf.for %arg2 = %c0_i64 to %10 step %c1_i64  : i64 {  // FAILED TO RAISE
      ...
    }
    ...
    %53 = arith.ori %arg1, %52 : i64
    affine.yield %53 : i64
  }

```

This is blocking #91.

julfarn:
Resolves #107


mlir/test/tools/qcc-opt/AffineRaise/scf-to-affine.mlir
```diff
@@ -55,9 +55,8 @@ func.func @nested_loop(%arg0: memref<?x?xi32>, %ub1: index, %ub2: index) {
 // CHECK-SAME:  %[[LB:.*]]: i32, %[[UB:.*]]: i32
 // CHECK:         %[[LB_IDX:.*]] = arith.index_cast %[[LB]] : i32 to index
 // CHECK:         %[[UB_IDX:.*]] = arith.index_cast %[[UB]] : i32 to index
-// CHECK:         affine.for %[[IV:.*]] = 0 to #{{.*}}%[[LB_IDX]]{{.*}}%[[UB_IDX]]
-// CHECK:           %[[IDX:.*]] = affine.apply #{{.*}}%[[LB_IDX]]{{.*}}%[[IV]]
```
rainij, line 59: I wonder why the step normalization is gone now. I did not look into the code so far. It makes actually more sense as it is right now as the step is known to be 1 here. I guess the step normalization kicked in due to a sub-optimal interaction with the index casting, right?

Just want to understand. Might be relevant for upstream.

I guess the other test is just affected the same way, right?

rainij: I think I understand now: So in the index cast step you "re-materialize a constant as an index constant" (from src) and this is what brings us into the static step case later on. The original code instead just inserted an index cast and that is not recognized as a constant anymore (by this `m_Constant` helper).

Is this correct?

rainij (commented):
LGTM. Just a small question, after that I will most likely approve.

mlir/test/tools/qcc-opt/AffineRaise/scf-to-affine.mlir
julfarn: Yes, exactly. If this was entirely indices instead of `i32` values, we would not expect (trivial) step normalization here either. This is now mirrored in the `i32` case. 

Instead of casting the constant `i32`, a fresh `arith.constant 1 : index` is inserted, and the old one is DCE'd.

rainij (approved):
Nice!

rainij:
@burgholzer can you have a quick look, this would unblock Julian's other PR which then finally closes the last issue of our first milestone.

burgholzer (approved):
just had a quick look and this looks reasonable

PR #112, opened 2026-06-30
IQPE integration test

### LINKED ISSUE #91: Qrisp to QIR for IQPE
Add an integration test for the iterative quantum phase estimation (qrisp to QIR to `qir-runner`). See also [MQT Bench](https://github.com/munich-quantum-toolkit/bench/blob/main/src/mqt/bench/benchmarks/iqpe.py). It should be added to the qrisp integration test framework (soon to be merged: #81).

Here is a qiskit based reference. Generated by claude and then simplified a bit. In this toy example the unitary is simply the phase gate $P(\phi)$ where $\ket{1}$ is the eigenvalue and $\phi=0.625$ is the "secret" value to to found by IQPE.

```python
from math import pi
from qiskit import QuantumCircuit, QuantumRegister, ClassicalRegister

phi_true = 0.625 # = 0.101 (binary) -> expect phi_1 phi_2 phi_3 = 1 0 1
m = 3            # bits of precision

sys = QuantumRegister(1, "psi")   # eigenstate qubit
anc = QuantumRegister(1, "a")     # the single ancilla, reused each iteration
cl  = ClassicalRegister(m, "phi") # cl[k-1] will hold bit phi_k
qc  = QuantumCircuit(sys, anc, cl)

qc.x(sys)  # prepare |1>, eigenstate of P(phi): P(phi)|1> = e^{i*phi}|1>

for k in range(m, 0, -1):  # LSB first: k = m, m-1, ..., 1
    qc.reset(anc)
    qc.h(anc)
    # controlled-U^{2^{k-1}}:
    qc.cp(2 * pi * phi_true * 2**(k - 1), anc[0], sys[0])

    # classical feedback: subtract phase of already-measured lower bits phi_{k+1..m}:
    for j in range(k + 1, m + 1):
        with qc.if_test((cl[j - 1], 1)):
            qc.p(-2 * pi * 2**(k - 1 - j), anc[0])

    qc.h(anc)
    qc.measure(anc[0], cl[k - 1])  # measure and store phi_k
```

julfarn:
Resolves #91


mlir/lib/Conversion/AffineRaise/SCFToAffine.cpp
rainij, line 71: This diff confuses me. Wasn't this update to `intBoundsRaisable` already merged. Can you please have a look? Looks like a version control issue.

mlir/test/tools/qcc/qrisp-integration/iqpe.py
```diff
@@ -0,0 +1,37 @@
...
+
+    measured_digits, *_ = q_fori_loop(
+        lower=0,
+        upper=precision,
+        body_fun=get_digit,
+        init_val=(measured_digits, phase_digit, target_phase, eigenstate, precision)
+    )
+
+    exit_code = q_cond(target_phase == measured_digits / (2**precision), lambda : 0, lambda : 1)
+    return exit_code
```
rainij: I would really prefer to output the measured digits here (it is just an integer right?). I stumbled over the check lines in the corresponding test. It looked really weird over there. I even feared that nothing is tested at all.

rainij (changes requested):

mlir/lib/Conversion/AffineRaise/SCFToAffine.cpp
julfarn: I think github didn't update its diff display. I clicked update Branch and now it seems gone.

mlir/test/tools/qcc/qrisp-integration/iqpe.py
julfarn: I wanted to justify this but then I noticed that I got confused: qrisp programs don't typically return an exit code. But the corresponding MLIR function will return an exit code, and the qrisp return gets transformed into output recordings. Will change.

julfarn: changed, please have a look

rainij (approved):

mlir/test/tools/qcc/qrisp-integration/iqpe.py
```diff
@@ -0,0 +1,40 @@
...
+
+    phase_digit = QuantumVariable(1)
+    measured_digits = 0
+
+    def get_digit(i, iter_args):
+        measured_digits, phase_digit, target_phase, eigenstate, precision = iter_args
+
+        reset(phase_digit[0])
+        h(phase_digit[0])
+        cp(2.0 * pi * target_phase * 2.0**(precision - i - 1), phase_digit[0], eigenstate[0])
```
burgholzer:
```suggestion
        cp(pi * target_phase * 2.0**(precision - i), phase_digit[0], eigenstate[0])
```
Just to simplify the computation

mlir/test/tools/qcc/qrisp-integration/iqpe.py
```diff
@@ -0,0 +1,40 @@
...
+
+        for j in jrange(i):
+            with control(((measured_digits >> j) & 1) == 1):
+                rz(-2.0 * pi * 2.0**(j - i - 1), phase_digit[0])
```
burgholzer:
```suggestion
                rz(-pi * 2.0**(j - i), phase_digit[0])
```

burgholzer (approved):
Just two minor simplifications. Otherwise this LGTM

mlir/test/tools/qcc/qrisp-integration/iqpe.py
julfarn: Sometimes I prefer to leave such calculations expanded, if it makes it easier to follow the calculation. (Factor 2 at the beginning comes in from a different source than the -1 in the exponent, they just happen to cancel out.) 

But I also agree that it's good to have a simple form here.

PR #116, opened 2026-07-10
HiSEPQ Loader

### LINKED ISSUE #67: Suggest opcode simulator output format
On the HiSEP-Q side we will soon (mid June) have a way to compile a test bench into a binary which can run on a generic linux host, like so:

```shell
# on my linux (freshly installed)
$ compiled_test_bench memory.mem
... some output
```

See https://github.com/caps-tum/HiSEP-Q-2.0/issues/4 .

At the start the output will be some arbitrary logging. The task is to devise a sensible output format so that the binary can be truly considered a "opcode simulator" in some sense.

This has to be done in coordination with the HiSEP-Q team and finally they have to implement the corresponding text bench. We on the other hand could in principle consider integrating the binary in CI testing (if this makes sense).

### LINKED ISSUE #69: Implement HiSEP-Q "Loader"
Within our LLVM fork implement a loader which takes the ELF produced by our compiler and generates the memory file for the opcode simulator (c.f. https://github.com/FullStaQD/compiler/issues/67). The memory file will look like this:

```
06600313
05500393
00400293
0C62F057
...
@00000400
06040200
00000000
...
```

The convention is that the top of the file corresponds to offset zero and contains the text section (PC starts at zero). In the above example at `0x400` the segment with the data section begins, but this is not fixed, it could be elsewhere.

In order to respect the convention above the linker has to be set up correctly. This probably amounts to devising an appropriate linker script. Once this is done the "loader" does not have to be smart or assume anything, it just grabs the info from the ELF. It is just that the ELF should specify that the text section should go to the very first memory segment.

### LINKED ISSUE #90: [HiSEP-Q] What to do about QIR-like _record_ operations?
Currently there is no obvious equivalent in HiSEP-Q. We have to tackle this problem with them together.

EnriqueNB:
Implements a linker loader script that use LLVM linker `lld` to link the output `.o` files into `.elf` files. For convenience, the loader script is embedded as a C++ header so it doesn't need to be called externally by the user. Moreover, an `elf2mem` tool has been created to convert the `.elf` to a `.mem` file suitable to be simulated by HiSEP-Q Verilator simulator. The options to compile to both `.elf` and `.mem` files have been added to `qcc`. 

`$ ./qcc --target=hisep-q --compile-to=mem ghz.mlir -o ghz.mem` 
`$ ./qcc --target=hisep-q --compile-to=elf ghz.mlir -o ghz.elf` 

This merge request tackles issue #69 and depends on MR #98.


EnriqueNB:
The loader is implemented. This MR depends on MR #98 and should be merged after it.

rainij (changes requested):
I cannot review this PR, it contains changes which I just reviewed for this other PR #98. Please re-request once the other one is merged or factor out the changes specific to this PR.

EnriqueNB:
To ease review I created a stack MR #123 to MR #98

Issue #115, opened 2026-07-08
[HiSEP-Q] Add custom LLVM Fork to CI


EnriqueNB:
Currently the CI for MR #98 doesn't work because it depends on our custom LLVM branch to be built. We need to incorporate in the CI a way to build the project with the custom LLVM branch.

julfarn:
To elaborate, we discussed the following:

Possible Goals: (Need to make decisions here)
1. We should be able to automate testing of all features, including those requiring modified LLVM. 
2. We should support building against a (specific, but official) LLVM version. 

As of #98, our compiler depends explicitly on our LLVM fork. It does not build for standard LLVM.
To meet goal 2, we would have to introduce some kind of featurization, and granulate our cmake targets so that a build against standard LLVM is possible once more (without the ability to lower to HiSep-Q).

For the CI, if we decide to support both LLVM kinds, there's no way around running everything twice. In the short term, I could see us simply disabling the problematic tests in the CI, so that progress can be made on #98.

Alternatively, committing completely to our own LLVM branch is a big decision. The large build will make it less accessible to users that may simply want to try out the software and use it to lower to QIR. Perhaps we can look into ways to patch standard LLVM, I don't know if there are techniques that could help us here. 

For now I would propose to 

1. Ensure it's possible to build a cmake target without HiSEP-Q functionality, that does not depend on our LLVM fork. (within #98)
2. Ensure the CI uses that cmake build config and skips tests that can't work. (within #98)
3. In a separate PR, address the gap in the CI coverage. 

Alternatively, we could take step 3 first, and modify the CI to run (additionally) on our fork, as suggested in the issue description. In any case this should happen in a dedicated PR.

burgholzer:
Just my 2 ct's:

I also believe that we should not require every potential user or contributor to a particular LLVM version; let alone a fork of LLVM. Especially so, if it is only for one of the (still rather niche) components of the compiler.
Thus, making that part optional as part of the CMake configuration should be the highest priority in my opinion. This should be as simple as guarding the respective code behind a CMake option.
I would further argue that this option should default to `OFF`, which also solves point two from above regarding CI only exercising paths of the code that it can reasonable exercise at the moment.

Working on point 3 then requires a bit more work.
One potential option could be that we also host pre-built distributions of selected commits of the fork in our `portable-mlir-toolchain` repository. This is only really feasibly if the fork has stabilized enough so that a new version isn't required every other day.
[...]

PR #117, opened 2026-07-13
cmake flag QCC_ENABLE_HISEP_Q=ON/OFF

rainij:
Relates to #115.

- Added a cmake flag to toggle support for the HiSEP-Q target (stub so far)
- Added a script to provision a patched version of LLVM in order to build against our custom backend.


mlir/lib/Target/QIR/QIRTarget.cpp
```diff
@@ -31,7 +31,7 @@
 
 namespace qcc {
 
-void buildQuantumPipeline(mlir::PassManager& pm) {
+void buildPipelineQIR(mlir::PassManager& pm) {
```
EnriqueNB: This implementation mix two different types pipelines: the hardware agnostic and the hardware aware one. The Jasp lowering, loop unrolling, and add static qubits passes are independent to building the QIR pass. The QIR pipeline should only consists of the passes related to QIR:
```
pm.addPass(qcc::createPrepToQIR());
pm.addPass(qcc::createConvertQCToQIR());
pm.addPass(qcc::createFinalizeToQIR());
```

After that you could add the passes according to the target. I propose to keep the general build quantum pipeline function and build the pipeline depending on different compilation options. As an example see the code in the prototype MR #116 

```
void buildQuantumPipeline(mlir::PassManager& pm, const Platform& platform) {
  const QuantumTarget& device = platform.getQuantumTarget();

  // Hardware agnostic: whatever the platform, a circuit of qc ops on statically indexed qubits.
  addJaspLoweringPasses(pm);
  addLoopUnrollingPasses(pm);
  addStaticQubitPasses(pm);

  // Device aware, while the circuit is still in the qc dialect: gate decomposition, qubit mapping
  // and QEC encoding.
  device.addDevicePasses(pm);

  // Controller aware: the whole way down from the circuit to the code of the control hardware. How
  // it gets there is its own business -- through QIR, or not (see `ControlTarget::addLoweringPasses`).
  platform.getControlTarget().addLoweringPasses(pm);
}
```

This is more future proof because we may even consider different hardware agnostic path (like the lowering from cuda-q we are considering now).

mlir/tools/qcc/qcc.cpp
```diff
@@ -140,8 +163,7 @@ int main(int argc, char** argv) {
     return 1;
   }
 
-  // TODO: Assemble pipeline based on target once a second valid target is added.
-  qcc::buildQuantumPipeline(pm);
+  targetInfo->buildPipeline(pm);
```
EnriqueNB: As said before, this doesn't take into account all possibilities of targets and front-end combinations so I would at least maintain the TODO and consider my previous comment.

mlir/include/qcc/Target/TargetRegistry.h
```diff
@@ -0,0 +1,38 @@
...
+#pragma once
+
+#include <functional>
+#include <llvm/ADT/ArrayRef.h>
+#include <llvm/ADT/StringRef.h>
+#include <mlir/Pass/PassManager.h>
+
+namespace qcc {
+/// Describes a compilation target selectable via `qcc --target=<name>`.
+struct TargetInfo {
```
EnriqueNB: I find the naming to be weird. I would consider this to be a target and I would rename the structure like so.
```
struct Target
```

mlir/include/qcc/Target/TargetRegistry.h
```diff
@@ -0,0 +1,38 @@
...
+  /// The `--target` value, e.g. "qir".
+  llvm::StringRef name;
+  /// Human-readable description shown by `--list-targets`.
+  llvm::StringRef description;
+  /// Whether this backend is compiled into the current build.
+  bool available;
```
EnriqueNB: Another useful field would be native quantum gates. If you want we can add it already or we can leave it to another MR, your choice.

mlir/include/qcc/Target/QIR/QIRTarget.h
```diff
@@ -0,0 +1,23 @@
...
+// ===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include <mlir/Pass/PassManager.h>
+
+namespace qcc {
+
+/// Assembles the lowering pipeline for the QIR target.
+void buildPipelineQIR(mlir::PassManager& pm);
```
EnriqueNB: In the line of other comments, I would change this to `addQIRPasses` so that we can have a more modular approach to building the pipeline.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -0,0 +1,26 @@
...
+//
+// TODO: For now this is a stub, add actual pipeline ASAP.
+//
+// ===----------------------------------------------------------------------===//
+
+#include "qcc/Target/HiSEPQ/HiSEPQTarget.h"
+
+namespace qcc {
+
+void buildPipelineHiSEPQ(mlir::PassManager& /*pm*/) {
```
EnriqueNB: I would reconsider this to be `addHiSEPQ` passes. See comment below.

mlir/lib/Target/HiSEPQ/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+# TODO: The HiSEP-Q backend is the only component that may depend on the HiSEP-Q
```
EnriqueNB, line 1: This TODO is cryptic to me. It is clear that HiSEP-Q depends of itself.

mlir/include/qcc/Target/TargetRegistry.h
```diff
@@ -0,0 +1,38 @@
...
+  /// Assembles the lowering pipeline for this backend. Only valid (non-null)
+  /// when `available` is true.
```
EnriqueNB: I would change it to `addBackendPasses` as said in other comments.

mlir/include/qcc/Compiler/Compiler.h
```diff
@@ -6,16 +6,18 @@
...
 #pragma once
-#include <mlir/IR/DialectRegistry.h>
+
 #include <mlir/Pass/PassManager.h>
 
 namespace qcc {
 
-/// A compilation pipeline assuming qc dialect as input and providing QIR-MLIR as output.
-///
-/// NOTE: The exact shape of the pipeline is still under construction.
```
EnriqueNB, line 18: I would keep this buildQuantumPipeline. See other comments.

EnriqueNB (commented):
I find the implementation of the patches correct for the current state of the codebase. It is a little bit annoying and not ideal but it is the standard solution and we have to bear with it. 

For the implementation of the targets. Conceptually, it is on the same conceptual line as mine in PR #116 so it is good for compatibilities reasons. I used classes for implementing the Target but I agree that for now structs are enough. That being said, I would rename `TargetInfo` to `Target` as I found this name to fit better the intend.

The only big difference that I see is that you eliminate completely the `buildQuantumPipeline` function and added a whole `buildTargetPipeline`  for each target. I consider this to not be future proof and conceptually wrong. Each target should only be its corresponding part of the pipeline and the entire pipeline should be built from the pipeline of each target (and possibly future frontends, quantum targets, etc).

Some add-ons like native gates associated to targets and the division between a quantum target and a control target can also be considered but we can move this to other pull requests and issues.

mlir/include/qcc/Target/TargetRegistry.h
rainij: Renamed to `Target`.

mlir/include/qcc/Target/TargetRegistry.h
rainij: I would not do it right away. We need a concept first. For example we mentioned that we want QDMI support in the future, it has methods to query this (see e.g. [QDMI_Operation](https://munich-quantum-software-stack.github.io/QDMI/v1.3.2/types_8h.html#ae777e8f92186c1e6f836eeaa53b149d7)) which would compete with this one.

mlir/lib/Target/HiSEPQ/CMakeLists.txt
rainij: I deleted it, wasn't a very useful comment anyway. But it means that the hisep-q target (inside QCC) is the only code which is allowed to depend on the code from the fork. This allows us to actually implement the `OFF` switch on the cmake side for HiSEP-Q.

mlir/lib/Target/QIR/QIRTarget.cpp
rainij: Yes I think you are right. I will change it mostly as you describe. I will add `buildPipeline(pm, target)` (just different name), and differentiate between target agnostic and target dependent passes. The individual targets will get this `addPasses(pm)` method. They assume that we start from QC, QCO and certain standard MLIR dialects. That makes sense.

I see a future need to also have passes depending on topology, routing etc, but I would keep it out for now (and also the near future). It would be just a dummy for now since we have no concept of e.g. topology at the moment. So there is no gain in adding it now. It also touches QDMI realm, so it needs some planning to get it right. Currently the mental model is that we assume an all-to-all connected device.

rainij: Now it is done. I also removed the "available" flag, I figured it is not worth the hassle and simplifies code alot.

mlir/tools/qcc/qcc.cpp
rainij: I now use `qcc::buildPipeline(pm, target);`. I guess it is better now without a comment. There is a comment in `TargetRegistry.h` explaining the design rationale and generally acknowledging the fact that this might get augmented in the future (with a hint to follow LLVM's lead).

mlir/include/qcc/Target/QIR/QIRTarget.h
rainij: I essentially did this (`addLoweringPassesQIR`).

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
rainij: Essentially did this (similar to QIR).

mlir/include/qcc/Target/TargetRegistry.h
rainij: Done (just with another method name).

mlir/include/qcc/Compiler/Compiler.h
rainij: It is back now as `buildPipeline(pm, target)`.

rainij:
@EnriqueNB thanks for your review, I believe I addressed all your requests. I essentially implemented it as you wanted with just minor deviations. Please have another look.

EnriqueNB (approved):
Looks good to me. Thank you for the implementation. Approved from my part.

mlir/include/qcc/Target/TargetRegistry.h
EnriqueNB: I will create an issue related to this so we can start shaping this up. Right now, in the HiSEP-Q backend we support a limited number of gates and the compiler should now about this. Moreover, soon we will start implementing hardware aware passes so it is a good moment to start this discussion.

mlir/lib/Target/QIR/QIRTarget.cpp
EnriqueNB: Perfect, thank you for the changes. When doing the review I thought about if having the "available" flag was worth the hassle. Good that we have a response now.

EnriqueNB:
@burgholzer Could you take a look? This is blocking MR #98 and we would like to have it merge asap.

rainij:
I merged it due to urgency (blocking other work by @EnriqueNB). If something needs to be addressed we can do it in one of the follow ups.

burgholzer:
> I merged it due to urgency (blocking other work by @EnriqueNB). If something needs to be addressed we can do it in one of the follow ups.

👍🏻 sorry for not coming back to this earlier. Had a busy week and am now on vacation until Monday. Quickly browsed through and this generally LGTM 👍🏻

PR #118, opened 2026-07-15
Simplify RUN commands by piping into qir-runner

julfarn:
Version 0.9.6 of `qirrunner` now supports reading QIR text from stdin. 

Only downside of the change is that now the compilation is run twice, so performance might be mildly worse. I don't think this is an issue now. Since we did encounter nondeterministic behavior previously it might even pay off.


julfarn:
We agreed not to make this change

PR #120, opened 2026-07-23
Add QISA CI workflow (QCC_ENABLE_HISEP_Q=ON)

### LINKED ISSUE #115: [HiSEP-Q] Add custom LLVM Fork to CI

rainij:
The primary CI builds qcc without QISA support against a prebuilt MLIR. The HiSEP-Q target needs a patched LLVM that no prebuilt package provides, so this new self-contained workflow builds patched LLVM from source (mlir only, Release + assertions), builds qcc with QCC_ENABLE_HISEP_Q=ON, and runs the lit suite including the `REQUIRES: hisep-q` tests.

The LLVM install prefix is cached keyed on base.rev and the QISA patches, so warm runs skip the LLVM build entirely; a ccache layer additionally speeds up rebuilds after an LLVM version or patch bump.

Resolves #115.

Assisted-by: Claude Code

### DIFF
--- a/.github/workflows/qisa-ci.yml
+++ b/.github/workflows/qisa-ci.yml
@@ -16,7 +16,7 @@ permissions:
 
 env:
   # Bump this to invalidate the LLVM install cache when the LLVM CMake flags
-  # below change (the cache key only tracks base.rev and the QISA patches).
+  # below change (the cache key does not track them).
   LLVM_CACHE_VERSION: v1
   # Do not set ccache max size too high. Total cache limit at the time of
   # writing this is 10GB. Note that LLVM install cache competes with this one.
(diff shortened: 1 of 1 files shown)


rainij:
Some data:
- Building LLVM takes a little below 3h.
- Restoring from cache takes around 20 seconds.
- Building QCC takes around 6 minutes, everything else is negligible.

rainij:
Ping @EnriqueNB

EnriqueNB (approved):
Looks good to me. Approved.

burgholzer (approved):
Looks reasonable to me 👍🏻

PR #123, opened 2026-08-03
Hisepq loader - to llvm-intrinsics

### LINKED ISSUE #69: Implement HiSEP-Q "Loader"

EnriqueNB:
- Implement `hisepq-elf2mem` to convert a linked binary to a verilog memory image
- Added linker script `hisepq.ld` with README on usage.
- Extended the existing lowering to HiSEP-Q to have a full running pipeline.
- Added corresponding tests.

Assisted-by: Claude Code


rainij:
I take over the PR because @EnriqueNB is in holiday's from tomorrow.

rainij (commented):
Review not needed as I work on this PR soon.

rainij:
I just rebased it onto main. I had to do this because it was not a clean stack and the stack workflow heavily choked on it (incomprehensible diffs etc). After some git wizardry it should be OK now.

rainij:
Went over it now, looks very good so far. I added a few FIXMEs on which I will work next week.

rainij:
@burgholzer finally ready. Please review :slightly_smiling_face:.

EnriqueNB (commented):
The new changes looks good to me. I specially like that adding __start is now a pass.

burgholzer (approved):
Only browsed through here. Tests are passing, and nothing directly popped out to me. Feel free to merge 👍🏼

PR #125, opened 2026-08-05
Update jasp

rainij:
- Update generator script to emit jasp in generic format (for stability reasons)
- Regenerated corresponding qrisp integration tests
- Added README with some background info into jasp folder

### DIFF
--- a/mlir/include/qcc/Dialect/Jasp/IR/README.md
+++ b/mlir/include/qcc/Dialect/Jasp/IR/README.md
@@ -1,0 +1,15 @@
+# Jasp IR
+
+The jasp dialect is not owned by us. We have copied it from
+[this folder](https://github.com/eclipse-qrisp/Qrisp/tree/main/src/qrisp/jasp/mlir/dialect_definition)
+from within the qrisp repository.
+
+In order to upgrade copy the files over here and possibly add include guards (or pragma once).
+
+TODO: Agree on a single source of truth for the dialect definition.
+See also issue https://github.com/FullStaQD/compiler/issues/16.
+
+Note that qrisp defines the dialects twice: On the one hand in the tablegen files already mentioned.
+Then on the other hand in xdsl (see e.g.
+[xdsl_dialect.py](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/jasp/mlir/xdsl_dialect.py).
+The former is needed for JAX, the latter they use to implement passes in python without the full MLIR dependency.
--- a/mlir/include/qcc/Dialect/Jasp/IR/README.md
+++ b/mlir/include/qcc/Dialect/Jasp/IR/README.md
@@ -4,7 +4,7 @@ The jasp dialect is not owned by us. We have copied it from
 [this folder](https://github.com/eclipse-qrisp/Qrisp/tree/main/src/qrisp/jasp/mlir/dialect_definition)
 from within the qrisp repository.
 
-In order to upgrade copy the files over here and possibly add include guards (or pragma once).
+In order to upgrade, copy the files over here and possibly add include guards (or pragma once).
 
 TODO: Agree on a single source of truth for the dialect definition.
 See also issue https://github.com/FullStaQD/compiler/issues/16.
--- a/mlir/utils/generate_qrisp_mlir.py
+++ b/mlir/utils/generate_qrisp_mlir.py
@@ -11,8 +11,7 @@
 
 # /// script
 # dependencies = [
-#   "qrisp==0.9.6",
-#   "xdsl==0.59.0"
+#   "qrisp[xdsl]==0.9.6",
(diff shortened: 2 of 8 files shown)


mlir/utils/generate_qrisp_mlir.py
```diff
@@ -66,6 +65,43 @@ def _find_qrisp_function(module):
     return candidates[0]
 
 
+def _to_selective_generic_mlir(module) -> str:
+    """Serialize an xDSL module with only ``jasp`` ops in generic form.
+
+    Rationale: custom jasp assembly has shown to be inconsistent between MLIR
+    and xdsl (qrisp has to duplicate dialect definitions). Well-known standard
+    dialects are still printed in custom assembly form for readability.
```
julfarn, line 73: Not sure I agree with this. The purpose of having the integration tests is to see if we can read qrisp output. Here you are "fixing" qrisp output. This might make our tests more stable, but it dilutes what they are testing. 

At the very least I would like some perspective: Should this be removed once the qrisp-mlir interface is more stable? Should this become part of the qrisp frontend in qcc? Or should this be part of the qcc backend in qrisp? 

This "fixing" of qrisp output is also highly specific to the current bug. Is it really likely that the exact same thing will happen again? (assembly mismatch but IR agreement) You were happy that our tests actually caught this issue, and with this in place they might no longer.

mlir/include/qcc/Dialect/Jasp/IR/README.md
```diff
@@ -0,0 +1,15 @@
+# Jasp IR
+
+The jasp dialect is not owned by us. We have copied it from
+[this folder](https://github.com/eclipse-qrisp/Qrisp/tree/main/src/qrisp/jasp/mlir/dialect_definition)
+from within the qrisp repository.
+
+In order to upgrade copy the files over here and possibly add include guards (or pragma once).
```
julfarn:
```suggestion
In order to upgrade, copy the files over here and possibly add include guards (or pragma once).
```

mlir/include/qcc/Dialect/Jasp/IR/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+TODO: Agree on a single source of truth for the dialect definition.
+See also issue https://github.com/FullStaQD/compiler/issues/16.
+
+Note that qrisp defines the dialects twice: On the one hand in the tablegen files already mentioned.
+Then on the other hand in xdsl (see e.g.
+[xdsl_dialect.py](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/jasp/mlir/xdsl_dialect.py).
+The former is needed for JAX, the latter they use to implement passes in python without the full MLIR dependency.
```
julfarn, line 15: This should go into the issue

julfarn (commented):
Looks good, please have a look at my concerns regarding the patching of qrisp output.

mlir/utils/generate_qrisp_mlir.py
rainij: I do not see the problem here. We were never really interested in their custom assembly format. To me a stable exchange format matters more and the generic format is more stable. What we caught here is an inconsistency on the qrisp side and it is nice of us to report that, but IMO it isn't really our job.

IMO the honest fix on the qrisp side is to actually turn off custom assembly format from the tablegen files. Apparently they do not need it.

mlir/include/qcc/Dialect/Jasp/IR/README.md
rainij: Added it via comment: https://github.com/FullStaQD/compiler/issues/16#issuecomment-5204599426

mlir/utils/generate_qrisp_mlir.py
julfarn: Ok, then we need to communicate that our input format for qcc is precisely this, and that we do not make any guarantees towards supporting jasp in custom assembly format.

julfarn:
I will have a final look when the subbranch is merged.

rainij:
> I will have a final look when the subbranch is merged.

@julfarn this is actually not how stacked PRs work. Say you have branches B -> A -> main. Then the normal workflow is as follows

- First merge A into main.
- Github automatically retargets B -> main.
- Then merge B.

I believe there is a shortcut "merge all" but conceptually it still works as above.

The point of the stacked PR is to keep the diff small for review. Here in this example it is not absolutely needed, but it arose naturally as the license issue just came to my mind when I already finished work here. If I would now merge B into A then this would technically invalidate reviews on A and make the diff bigger.

julfarn:
Ah yes of course, this makes more sense in this situation. We had other cases in the past where we would merge smaller fixes into a larger feature branch, I thought this was what you were going for here.

julfarn (approved):

PR #126, opened 2026-08-06
Resolve jasp license issue

rainij:
We wrongly slapped our license header on the two tablegen files we copied from qrisp.

- Remove our license header from jasp, move to third-party/.
- Move the corresponding README and improve it.


julfarn (approved):
Reasonable to have a dedicated third party directory. Good catch with the licenses.

rainij:
> Reasonable to have a dedicated third party directory. Good catch with the licenses.

I believe that I actually added those headers when I bulk added them to every file. I just overlooked those two files in the sea of files which were changed. Again an example to be extra-careful with bulk edits.

burgholzer (approved):
LGTM

PR #134, opened 2026-08-24
Added QVec Dialect

rainij:
- Added new dialect `qvec` and corresponding interfaces.
- Added pass to lower `qco` to `qvec`.
- Added pass to merge vector ops to ops acting on more qubits.

This is just the start; follow up work will extend and improve it. QCC integration will also come soon.

Assisted-by: Claude Code


mlir/tools/qcc/CMakeLists.txt
```diff
@@ -25,7 +25,25 @@ set(LIBS
...
+  set(RESCAN_SUPPORTED ${CMAKE_LINK_GROUP_USING_RESCAN_SUPPORTED})
+endif()
+
+set(LINK_LIBS ${LIBS})
+if(RESCAN_SUPPORTED)
+  list(REMOVE_ITEM LINK_LIBS MLIRQCODialect MLIRQTensorDialect MLIRQTensorUtils)
+  list(APPEND LINK_LIBS "$<LINK_GROUP:RESCAN,MLIRQCODialect,MLIRQTensorDialect,MLIRQTensorUtils>")
+endif()
+
+target_link_libraries(qcc PRIVATE ${LINK_LIBS})
```
rainij, line 46: @burgholzer this is an AI generated fix for a build failure I had after the MQT update (only this branch for some reason). Sorry for throwing it over the fence without me verifying it first, but I have to finish the day early today. Is the issue known already? Or is the claim of the AI wrong (the circular dependency between QCO and QTensor)?

burgholzer: This sounds wrong to me at first glance. The QCO dialect has no dependence on the QTensor dialect.
I am afk until tomorrow afternoon. I'll take a more detailed look over the weekend.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
```diff
@@ -0,0 +1,352 @@
...
+  SmallVector<Operation*> members;
+  /// Total number of qubits per operand slot, i.e. the VF the merged operation would have.
+  int64_t width = 0;
+  /// The qubits the group acts on, i.e. the merged operations would have.
+  DenseSet<Value> qubits; // FIXME: TypedValue of Qubit?
+
+  /// Build the first (slot=0) or second (slot=1, if available) qubit vector operand for the merged operation by
+  /// collecting each member's qubits in the same slots.
+  Value buildMergedOperand(OpBuilder& builder, Location loc, unsigned slot) const {
+    assert(slot == 0 || slot == 1 && "slot can only be 0 or 1");
```
EnriqueNB, line 121:
```suggestion
    assert((slot == 0 || slot == 1) && "slot can only be 0 or 1");
```

There is a missing parenthesis as the && has priority over the ||.

mlir/include/qcc/Conversion/ToHiSEPQ/ToHiSEPQ.td
```diff
@@ -75,3 +75,54 @@ def EmitHiSEPQStart : Pass<"emit-hisepq-start", "mlir::ModuleOp"> {
...
+      - Every qubit vector traces back to a `vector.from_elements` whose elements are all `qco.static` ops, so that the
+        qubit indices are known at compile time.
+        TODO: Lift this once the dialect grows dynamic qubits. Then the qubit vector becomes a runtime value and the
+        constant folding here turns into a type conversion.
+      - Qubit indices are encoded in an unsigned `i{QEW}`, and there are no more qubits than LMUL 8 holds at the minimum
+        VLEN.
+
+      Gates without an intrinsic are rejected (error).
+
+      `llvm.riscv.qv.mz` returns void and the QISA specifies no way to read a measurement back, so the classical result
```
EnriqueNB, line 113:
```suggestion
      `llvm.riscv.qv.mz` returns void as the QISA doesn't specify how to read a measurement back, so the classical result
```

Small typo.

mlir/lib/Conversion/QCOToQVec/ConvertQCOToQVec.cpp
```diff
@@ -0,0 +1,183 @@
...
+  void runOnOperation() override {
+    ModuleOp moduleOp = getOperation();
+    auto* ctx = moduleOp.getContext();
+
+    ConversionTarget target(*ctx);
+    target.addLegalDialect<QVecDialect, vector::VectorDialect>();
+    target.addIllegalDialect<qco::QCODialect>();
+    target.addLegalOp<qco::StaticOp>(); // still needed as qubit source
+
+    // FIXME: Why? Is this even correct?
```
EnriqueNB, line 161: This is done so there is no conflict between the `CtrlLowering` and the `SingleGateLowering`. The pass lower the control gates by recognising them and then extracting the gate inside this would produce a conflict with `SingleGateLowering` because it would also recognise a gate inside the control and it would try to lower it. By declaring as legal the gates inside a control body, the problem is eliminated.

mlir/lib/Conversion/ToHiSEPQ/ConvertQVecToHiSEPQIntrinsics.cpp
```diff
@@ -0,0 +1,332 @@
...
+    return success();
+  }
+
+  Diagnostics* diags;
+  HiSEPQMachine machine;
+};
+
+/// Rewrites `qvec.mz` into `llvm.call_intrinsic "llvm.riscv.qv.mz"` plus a poison result.
+///
+/// TODO: The QISA specifies no way to read a measurement back, so the classical bits are lost
```
EnriqueNB, line 261: This is already implemented in HiSEPQ so we should add it.

EnriqueNB (commented):
Looks good to me. I added a few comments.

mlir/tools/qcc/CMakeLists.txt
burgholzer: Investigating this now. There is something to fix here for us in MQT Core. Will post an update that hopefully makes this workaround irrelevant.

burgholzer: The underlying issue is fixed by https://github.com/munich-quantum-toolkit/core/pull/2292 which just landed on `main` as https://github.com/munich-quantum-toolkit/core/commit/35566470c04c1acc3a4d0a00a5f0a2550ace82e0

mlir/lib/Conversion/ToHiSEPQ/ConvertQVecToHiSEPQIntrinsics.cpp
rainij: Wasn't when I started; will postpone to follow up issue. This PR motivated me to ask them to implement the support.

mlir/lib/Conversion/QCOToQVec/ConvertQCOToQVec.cpp
rainij: I actually removed this. I think it is a misuse and guards against something that cannot happen. I believe the reasoning of claude putting it there was that if the inner stuff is converted first this would make the ctrl op being impossible to convert (as you say). But this is a scheduling question, legalization should not handle this. I checked that conversion hardcodes pre-order and hence the ctrl op is converted first. They cannot change this without breaking everybodies code so this can be assumed a contract.

mlir/include/qcc/Dialect/QVec/IR/QVec.h
```diff
@@ -0,0 +1,84 @@
...
+#define GET_ATTRDEF_CLASSES
+#include "qcc/Dialect/QVec/IR/QVecAttrs.h.inc"
+
+//===----------------------------------------------------------------------===//
+// QVec type predicates
+//===----------------------------------------------------------------------===//
+
+namespace qcc::qvec {
+
+/// Whether `type` is a vector of qubits, the type every qubit slot carries.
```
julfarn: a bit confusing IMO, would go with something like

```suggestion
/// Whether `type` is a vector with element type `!qco.qubit`.
```

mlir/include/qcc/Dialect/QVec/IR/QVec.h
```diff
@@ -0,0 +1,84 @@
...
+//===----------------------------------------------------------------------===//
+// QVec Operations
+//===----------------------------------------------------------------------===//
+
+#define GET_OP_CLASSES
+#include "qcc/Dialect/QVec/IR/QVecOps.h.inc"
+
+namespace qcc::qvec {
+
+/// Trace back the qubit at `index` in the vector `qubits` to a StaticOp and return it if possible (null value if not).
```
julfarn: General question: We should decide if we stick to returning null in functions like these or making the return type `std::optional`. Right now we are a bit inconsistent. I generally favor making the possibility of a null return explicit.

julfarn: Also, docstring style: `Trace` here, `Adds` below.

mlir/include/qcc/Dialect/QVec/IR/QVecInterfaces.td
```diff
@@ -0,0 +1,109 @@
...
+
+//===----------------------------------------------------------------------===//
+// QubitSlotOpInterface
+//===----------------------------------------------------------------------===//
+
+def QVec_QubitSlotOpInterface : OpInterface<"QubitSlotOpInterface"> {
+  let description = [{
+        A qubit slot is one lane of qubits through an operation: a qubit vector result together with the operand that
+        carries the same qubits in (if the latter exists). A `qvec.single` has one slot, `0`; a `qvec.pair` has two, `0`
+        and `1`.
```
julfarn: Not proposing a change at this point, just remarking:

While you need this for the specific purpose of tracing back values through operations, the concept of this interface is much more general. It applies to all operations for which we can identify (perhaps partially) some input value "becoming" some output value. This kind of information seems important for allocation (going back to reference semantics). I guess it is usually implicitly encoded: 

If `%b0, %b1 = value.op %a0, %a1` gets lowered to `ref.op %ref0, %ref1`, this lowering rule **implies** that `%b0` came from `%a0` and `%b1` came from `%a1`. But it's nice to have it explicit, and it is really an enriching structure to the IR DAG.

mlir/include/qcc/Dialect/QVec/IR/QVecInterfaces.td
```diff
@@ -0,0 +1,109 @@
...
+#ifndef INCLUDE_QCC_DIALECT_QVEC_IR_QVECINTERFACES_TD_
+#define INCLUDE_QCC_DIALECT_QVEC_IR_QVECINTERFACES_TD_
+
+include "mlir/IR/OpBase.td"
+
+//===----------------------------------------------------------------------===//
+// QubitSlotOpInterface
+//===----------------------------------------------------------------------===//
+
+def QVec_QubitSlotOpInterface : OpInterface<"QubitSlotOpInterface"> {
```
julfarn: Wouldn't `QvecSlot...` be a better name as each slot is for a vector of qubits, not a qubit?

mlir/include/qcc/Dialect/QVec/IR/QVecOps.td
```diff
@@ -0,0 +1,144 @@
...
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+//===----------------------------------------------------------------------===//
+// QVec operations
+//===----------------------------------------------------------------------===//
+
+// TODO: The memory effects we use here are probably too conservative. We have them here to protect our ops against
+// quantum-unaware passes. Note that e.g. a `Pure` measurement would allow some passes to hoist them out of control-flow
+// or alternatively sink into control-flow (see code motion) under circumstances were this is not consistent with the
+// intended semantics.
```
julfarn, line 28: Could we discuss an example? (Not necessarily here)

I'm quite interested. 

> not consistent with the intended semantics

Does this mean incorrect compilation in the sense that the resulting program behaves unexpectedly (as a black box), or only unintended compilation in the sense that the resulting program performs unintended steps (e.g. unnecessary measurement), but produces correct outputs?

mlir/include/qcc/Dialect/QVec/IR/QVecOps.td
```diff
@@ -0,0 +1,144 @@
...
+        and yields the measured qubits plus one classical bit per element.
+
+        Example:
+        ```mlir
+        %qs_out, %r = qvec.mz %qs : vector<4x!qco.qubit> -> vector<4xi1>
+        ```
+    }];
+
+  let arguments = (ins QVec_QubitVector:$qubits_in);
+  let results = (outs QVec_QubitVector:$qubits_out,
```
julfarn, line 121: Out of curiosity: Say you didn't return the qubits here and only the bits. Would there be any problems? You would have zero slots, but you also wouldn't need to do any backtracing. Would anything actually break?

mlir/lib/Dialect/QVec/IR/QVecDialect.cpp
```diff
@@ -0,0 +1,232 @@
...
+      >();
+
+  addOperations<
+#define GET_OP_LIST
+#include "qcc/Dialect/QVec/IR/QVecOps.cpp.inc"
+      >();
+}
+
+//===----------------------------------------------------------------------===//
+// Qubit provenance
```
julfarn, line 57: This is a nice word, I remember we had some trouble finding the right term and went with "backtracing" for the procedure which sounds too vague. 

Perhaps this could also inspire a new name for the slot interface?

julfarn (commented):
This is a first partial review in which I went over the dialect definition, but have not looked at the passes yet.

mlir/include/qcc/Dialect/QVec/IR/QVecOps.td
rainij: Pure operations can be moved around quite liberally. In the presence of control flow it can happen that the transformed program is just not equivalent to the original program. I can try to make a concrete example (in the sense of running MLIR) when we discuss the side effects topic in our tech meeting. Never tried it but quite certain from what I read about pure.

mlir/lib/Dialect/QVec/IR/QVecDialect.cpp
rainij: Claude used it already somewhere else earlier `third-party/qrisp/README.md`. I also like the word but it is a bit lengthy.

mlir/include/qcc/Dialect/QVec/IR/QVecInterfaces.td
rainij: You mean `QVec_QVecSlotOpInterface`? Reads a bit weird with the double QVec. The first one translates to a C++ namespace. The "qubits" should refer to the fact that we only care about the qubit operands. In principle this slot interface is pretty generic.

But because you talk about naming in another place. "Slot" is not set in stone. "Lane" is also an option I would consider: `QVec_LaneOpInterface`. Here lane (or slot) really just means, there are pairs of operands and results which correspond to each other. I keep the "qubit" in the name to say that it is about qubits. Would be hard to make sense without qubits, for example for a measurement the bits also correspond to the qubit operands (here the lane would duplicate if one applies it to non-qubit stuff too).

mlir/include/qcc/Dialect/QVec/IR/QVecOps.td
rainij: Technically yes because I designed the interface in an asymmetric way. `getQubitResult(slot)` is supposed to return something non-null while `getQubitOperand(slot)` is not (can be a value equivalent to null).

Could in principle be symmetrized. There is even a comment in the docstring of the interface itself talking about this. Currently we assume there are no qubit sinks (only sources and passthroughs). But one can question this. (Btw: not even sources exist right now, but it is plausible that we might need something like `qvec.alloc` or `qvec.static`). Your "destructive measurement" would be such a sink.

mlir/include/qcc/Dialect/QVec/IR/QVecInterfaces.td
rainij: I agree, the concept is really not `QVec` specific. Having multiple quantum dialects around (in value semantics) one likely needs it for each of them.

mlir/include/qcc/Dialect/QVec/IR/QVec.h
rainij: It is not really inconsistency. When returning an op it is idiomatic to not use an optional in general. Ops are like smart pointers. You have to check if they carry a value before usage, or rely on the promise of the guy who gave you the op (or value). But this doesn't mean one shouldn't use optional in other (non-pointer cases).

If one wants to put it into a rule: An op or `Value` is usually not wrapped into an optional.

julfarn: Ok that's fair, I'll stick to this convention

mlir/include/qcc/Dialect/QVec/IR/QVecInterfaces.td
julfarn: I think you are right, as it's a bit of a two-purpose interface. It allows you to ask the question "does this input and this output wire of type QVec share a  hardware ressource", but also "does this qubit living in the qubit input wire point to the same hardware qubit as the qubit living in this output qvec wire"? 

The first is quite general, the second more specialized. 

I like "lane" because you get the picture of a ressource passing through the op, which is accurate. No super strong feelings though

mlir/include/qcc/Dialect/QVec/IR/QVecOps.td
julfarn: Yes let's look at this in detail, I would be suprised if something that's legal classically would be illegal in the quantum context.

mlir/include/qcc/Dialect/QVec/Transforms/Passes.td
```diff
@@ -0,0 +1,57 @@
...
+      // Assuming %v0, %v1, %v2 are known to refer to disjoint sets of qubits.
+      %h0 = qvec.single h %v0 : vector<4x!qco.qubit>
+      %x1 = qvec.single x %v1 : vector<3x!qco.qubit> // despite obstacle we can still merge the hadamards
+      %h2 = qvec.single h %v2 : vector<2x!qco.qubit>
+      ```
+
+      turns into
+
+      ```mlir
+      %5 = qvec.single h %4 : vector<6x!qco.qubit>
```
julfarn, line 42: How is `%4` produced from `%v0` and `%v2`? This is more important here than showing that the pass can handle obstacles like the x-gate, which a user will expect anyways.

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
```diff
@@ -0,0 +1,255 @@
...
+    func.return
+}
+
+// CHECK-DAG:     %[[Q0:.*]] = qco.static 0
+// CHECK-DAG:     %[[Q2:.*]] = qco.static 2
+// CHECK:         %[[V0:.*]] = vector.from_elements %[[Q0]], %[[Q2]] : vector<2x!qco.qubit>
+// CHECK:         qvec.single h %[[V0]] : vector<2x!qco.qubit>
+// CHECK:         qvec.single x %{{.*}} : vector<1x!qco.qubit>
+
+// -----
```
julfarn: Is there a particular reason why you test for this? This, like most passes, only look at the graph structure of the IR and don't care about order of operations. It's not a particular feature of the pass that this works, and it would be more work to write the pass in such a way that it doesn't.

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
```diff
@@ -0,0 +1,255 @@
...
+
+// CHECK-LABEL: func.func @dependent_pairs
+func.func @dependent_pairs() {
+    %q0 = qco.static 0 : !qco.qubit
+    %q1 = qco.static 1 : !qco.qubit
+    %v0 = vector.from_elements %q0 : vector<1x!qco.qubit>
+    %v1 = vector.from_elements %q1 : vector<1x!qco.qubit>
+
+    %a, %b = qvec.pair cx %v0, %v1 : vector<1x!qco.qubit>
+    %c, %d = qvec.pair cx %v1, %v0 : vector<1x!qco.qubit>
```
julfarn: Double use of `%v0`, `%v1`. Is this legal IR? Doesn't matter that much for the test, but I think we should stick to testing on legal IR whenever possible.

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// Do not merge!
+// CHECK:         qvec.pair cx %{{.*}}, %{{.*}} : vector<1x!qco.qubit>
+// CHECK:         qvec.pair cx %{{.*}}, %{{.*}} : vector<1x!qco.qubit>
+
+// -----
+
+// CHECK-LABEL: func.func @interleaved_layers
+func.func @interleaved_layers() {
```
julfarn: Here I am also not sure whether the "interleaving" part of the test needs highlighting.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
```diff
@@ -0,0 +1,365 @@
...
+#include <algorithm>
+#include <cstdint>
+#include <limits>
+#include <utility>
+
+using namespace mlir;
+using namespace qcc::qvec;
+
+//===----------------------------------------------------------------------===//
+// A uniform view of the qvec operations
```
julfarn: This headline is confusing to me

mlir/tools/qcc/CMakeLists.txt
rainij: Yes, updated here too and it worked worked :).

mlir/include/qcc/Dialect/QVec/IR/QVecOps.td
rainij: I added this to my personal task on studying side effects. We will soon have a discussion on that.

mlir/include/qcc/Dialect/QVec/IR/QVecAttrs.td
```diff
@@ -0,0 +1,56 @@
+// ===----------------------------------------------------------------------===//
```
EnriqueNB, line 1: The standard practice is to put the enums in a separate file than the attributes(see for example https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Linalg/IR/LinalgEnums.td) . I would rename this file to `QVecEnums`.

EnriqueNB (changes requested):
The dialect successfully capture the vector necessities to lower to HiSEPQ. Looks good to me. I only suggest a small change.

mlir/include/qcc/Dialect/QVec/IR/QVecAttrs.td
rainij: Moved it to `QVecEnums.td`.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
rainij: Thanks, fixed it.

EnriqueNB (approved):

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
```diff
@@ -0,0 +1,365 @@
...
+    auto* ctx = moduleOp.getContext();
+
+    // A max-vf of 0 leaves VF (vectorization factor) unbounded; see the pass description.
+    const int64_t limitVF = maxVF == 0 ? std::numeric_limits<int64_t>::max() : maxVF;
+
+    SmallVector<Block*> blocks;
+    moduleOp->walk([&](Block* block) { blocks.push_back(block); });
+    for (Block* block : blocks) {
+      mergeOpsInBlock(*block, limitVF);
+    }
```
julfarn, line 348: I'm wondering if latching onto the module is the best approach here. This is highly parallelizable, but executed sequentially for each block. Is there a reason why you didn't schedule this on functions?

julfarn, line 348: We may want to justify the approach a given pass takes. I think it's quite a big decision to go for a custom walk or rewrite pattern approach. Here, the reason is that you want to connect possibly disconnected (in the IR graph sense) parts of the IR. This is inherently non-local, meaning you can't match on an op and modify its neighborhood to achieve this. Something like this would be nice to have documented, to build trust that a particular implementation is well-designed.

In other cases I remember we were worried that we used rewrite patterns where a custom walk would have been better.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
```diff
@@ -0,0 +1,365 @@
...
+/// interleaved in the IR).
+///
+/// *Grouping:* Within each bucket we form groups of `qvec` operations (meant to be merged in the end). A group is a
+/// list of `qvec` operations ordered as in the IR. Before a candidate is added to a group, the following tests have to
+/// pass:
+///
+/// 1. The enlarged group still respects `limitVF` once merged.
+/// 2. All qubits of the candidate can be traced back to static ops.
+/// 3. All new qubits are disjoint from the ones the group already operates on (a safety net against incomplete
+///    producers).
```
julfarn, line 237: Doesn't "all qubits can be traced back" imply that there are no incomplete producers anywhere preceding the op?

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
```diff
@@ -0,0 +1,365 @@
...
+/// *Producers:* The *producers* of a `qvec` op are all the `qvec` ops immediately involved in creating its qubit
+/// operands.
+///
+/// *Layering:* Every `qvec` op is assigned an integer layer: `layer(op) = 1 + max{layer(producer)}` over all producers
+/// of `op`. An op without producers (e.g. the first `qvec` op in the block) gets layer `0`.
+///
+/// IMPORTANT: We cannot always determine all producers (incomplete producers), for example because unknown ops "block"
+/// the way. Any pipeline this pass runs in is urged to avoid that scenario, as we optimize under the assumption of
+/// complete producers. An underestimated layer propagates downstream: an op whose own producers are complete still gets
+/// a wrong layer if any op upstream of it has incomplete ones.
```
julfarn, line 214: I'm getting the impression that removing all ops with incomplete producers as well as their decendents from consideration would simplify the logic of this pass. Have you considered this?

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+/// Identical bucket key for two ops expresses the fact that they can in principle be merged if no other op blocks and
+/// qubit slots are disjoint. BucketKey = (operation name, secondary bucket key).
+using BucketKey = std::pair<OperationName, uint32_t>;
+
+/// Get the second component of the `BucketKey`.
+///
+/// Returning nullopt at runtime is considered a bug, hence the assert in the impl. We still emit a null-opt to avoid
+/// correctness mistakes during production run. Returning a nullopt just means a possibly missed merge opportunity.
+static std::optional<uint32_t> getSecondaryBucketKey(QubitSlotOpInterface op) {
```
julfarn: I'd propose to rename this. This function takes an op and returns the gate type.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
```diff
@@ -0,0 +1,365 @@
...
+  assert(secondaryKey && "unhandled qvec operation, it stays unmerged");
+  return secondaryKey;
+}
+
+//===----------------------------------------------------------------------===//
+// Scheduling helpers
+//===----------------------------------------------------------------------===//
+
+/// Makes `value` available at `before`, hoisting whatever defines it if it is not already. Returns true iff succeeded.
+/// IR might still be mutated if unsuccessful (but still correct).
```
julfarn, line 69: I have seen the word "hoist" used for "hoisting" things out of loops. This refers to the nesting dimension of IR ordering. Here you use it to refer to the dominance/ordering dimension. I'd rather reserve the term for only one concept, unless there's precedent.

julfarn (commented):
I have now reviewed the merging pass.

Overall it seems well-structured and as complex as it needs to be. All my comments are minor or ask for clarification. 

However, I would like to take the opportunity to question the layering bookkeping once more and ask what we gain from it. Here's my understanding. 

Within a block, and not looking into subregions, the IR has the structure of a DAG (allowing multi-edges, but we can treat those as implicit copy ops, so no loss of generality), equipped with an additional linear order (dominance relation) that refines the partial order defined by the DAG. In the presence of side-effecting operations, the dominance relation carries meaning, otherwise it doesn't. 

In constructing the layer map, you create another ordering-like structure from **part of** the DAG: by looking only at slot-ops and slot-values, you get a sub-DAG of the original IR (obtained by deleting non-slot-ops (vertices) and non-slot values (edges)). This DAG again has a partial order, and your layers are compatible with this order in some precise-but-I-dont-want-to-spell-it-out sense. 

But the layer structure is not necessarily compatible with the partial order of the whole DAG. You account for this later by doing some checking on whether you can reorder IR. 

This split into two problems seems artificial to me. In the end, what you are checking for is really:
 - Does merging these two ops create a loop in my IR? (Violating the A in DAG.) You check this by first considering only ops in the same layer (guarantees no cycles of slot-values) and then by checking if you "can move this op before the other op" (guarantees no cycles at all). 
 - Does merging these two ops move a side-effecting op? Better not touch those.

I believe asking these two questions to every pair of ops in a block and merging greedily would result in the same behavior with less complexity of code (haven't thought about complexity of the algorithm). I don't think the slot interface is needed here, except for handling IR that we should consider illegal anyways, like
```
%qubit = qco.static 0
%same_qubit = qco.static 0
... // Nothing good can come out of this
```
It's a fair assumption (IMO the foundational principle of value semantics) that qubit values or qubit vector values that may live in parallel in the IR (no SSA DAG relation) can be thought of as living on disjoint physical qubit (sets). If we accept this, only looking at the DAG structure and paying attention to side-effecting ops should be enough. 

I'm aware "why can't we do this completely differently" is not a good review, and I don't want to block here. I'm convinced of the correctness of the current implementation. But maybe I can ask of you:
 - Your thoughts on this.
 - If you can (quickly) think of a counterexample where the algorithm I proposed does not work.
 - If not, open an issue to revisit this point at a later time.

mlir/include/qcc/Dialect/QVec/IR/QVec.h
rainij: Don't understand what you mean. Should I change the docstring?

mlir/include/qcc/Dialect/QVec/IR/QVecInterfaces.td
rainij: Makes sense, then I would tend to rename it to `QubitLaneOpInterface` and update all docstrings to use the new terminology etc. Do you think this is an improvement?

julfarn: Yes, let's do it that way!

mlir/include/qcc/Conversion/ToHiSEPQ/HiSEPQMachine.h
```diff
@@ -0,0 +1,79 @@
...
+  ///
+  /// NOTE: This implies that the type `vector<[N]xi{QEW}>` corresponds to the given LMUL.
+  [[nodiscard]] unsigned knownMinElementsFor(unsigned lmul8) const;
+
+  /// The number of bits LLVM treats as the known part of a scalable vector on RISC-V. Mirrors
+  /// `llvm::RISCV::RVVBitsPerBlock`, which we cannot include here because only the target layer
+  /// may depend on the HiSEP-Q LLVM fork.
+  static constexpr unsigned rvvBitsPerBlock = 64;
+
+  /// The LMUL register groups the QV instructions address, narrowest first, stored as eighth times
```
julfarn:
```suggestion
  /// The LMUL register groups the QV instructions address, narrowest first, stored as eight times
```
typo

mlir/include/qcc/Dialect/QVec/IR/QVecInterfaces.td
rainij: Done.

mlir/include/qcc/Dialect/QVec/Transforms/Passes.td
rainij: I added a comment explaining it at a high level. I did not write IR because it would only be complete if I listed the `qco.static` declarations and this would drown the actual content of the snippet.

mlir/lib/Conversion/ToHiSEPQ/ConvertQVecToHiSEPQIntrinsics.cpp
```diff
@@ -0,0 +1,331 @@
...
+    ModuleOp moduleOp = getOperation();
+    auto* ctx = moduleOp.getContext();
+
+    // A VLEN below `64` would make `vscale` a fraction, and anything that is not a power of two is not a VLEN at all.
+    if (minVLen < 64 || !llvm::isPowerOf2_32(minVLen)) {
+      emitError(moduleOp.getLoc()) << "'min-vlen' expects a power of two of at least 64, got " << Twine(minVLen);
+      return signalPassFailure();
+    }
+
+    // 8 and 16 are the only save values for QEW for our currently hardcoded set of possible LMUL values (see
```
julfarn:
```suggestion
    // 8 and 16 are the only safe values for QEW for our currently hardcoded set of possible LMUL values (see
```

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
rainij: I expect the pass to be improved in the future (currently it is not particularly good at finding merge opportunities). For each improvement I would like to add a test. This documents a tiny capability. I expect any reasonable implementation to satisfy this but it is possible that somebody reimplements it by a walk over the IR and greedy merges of subsequent gates. But I have no hard feelings, hard to judge which tests are really useful.

mlir/test/tools/qcc-opt/QCOToQVec/convert-qco-to-qvec.mlir
```diff
@@ -0,0 +1,123 @@
...
+// CHECK:         return %[[BIT]] : i1
+
+// -----
+
+// CHECK-LABEL: func.func @static_qubits_survive
+func.func @static_qubits_survive() {
+    %q0 = qco.static 7 : !qco.qubit
+    %h = qco.h %q0 : !qco.qubit -> !qco.qubit
+    func.return
+}
```
julfarn, line 121: Why do you only check for static, not the h gate here?

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
rainij: Good that you catched this, corrected it.

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
rainij: OK the second complaint on the tests :D. I think the reference to the "layer" algorithm is the problem - the tests should not care. But it is a nice semi-complex example so I moved it to the top of the file as the second test case (after an even simpler one). Called them "example_1" and "example_2". Serves now the reader seeking to understand what the pass does. I deleted the test from an earlier comment. I agree tests should be simple, this was my poor attempt to make it slightly better.

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
rainij: OK upon thinking, we have another test which tests this implicitly. Should be enough. Removed it.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
rainij: Removed it. Don't need a headline for everything.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
rainij: Not sure what you mean. This algorithm is really not expressible in terms of rewrite patterns - for obvious reasons. I see no point in justifying not using them.

That being said: I expect this algorithm or the implementation to be replaced by something more idiomatic. I do not like it at the moment, but I cannot point the finger on it. I cannot do it right now as I need further information to improve it. I intended it as a minimal implementation, the layer idea sounded super clean on paper when I came up with it, but it turned out to be not so clean upon implementation. The good thing though: this implementation is completely private to this file, so chances of dragging down other code is low.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
rainij: No. Let us start from the good case: If all producers are complete than one can skip check 3 always. But as soon as one op has an unknown producer all its descendents are prone to be in the wrong layer. One could either track this or introduce check 3.

The rationale is this: We expect the good case and try hard to ensure it is satisfied. But this can only ever be a best effort: There are parts of the tracking code which have to handle individual operations like `vector.from_elements`. If somebody adds an new op we are not aware the pass would fail. So in the bad case we try to be correct. There is more to it, we can talk about it if you like, I pondered a lot about this. Don't won't to write all my thoughts.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
rainij: Yes, but would make the pass potentially weaker (definitely not stronger) and still require additional analysis (currently there is no need to track the decendents). So unclear how much easier to understand it would be.

I think this is not really necessary to investigate. IMO with more insights we might be able to find a different approach which is better and easier to understand at the same time. I would do it in another PR. The main goal of this PR is to introduce the QVec dialect and prove that we can indeed translate qco to intrinsics. I would have left out the pass for readability but I wanted that proof.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
rainij: I do not know any better name which is not misleading. It computes the second component of the bucket key. So in a sense it fits. Let me know if you have a better suggestion.

julfarn (commented):
I have now reviewed the lowering passes. 

I did not go over the HISEP-Q part quite as thoroughly because it would have taken me a long time as I'm not an expert.  But it looks very reasonable and straightforward. 

My only question is: Is it correct to lower directly to assembly from a value-semantic dialect? Conceptually there's two steps: first going to reference semantics, then lowering to a particular backend representation. It may make sense like this, but my instinct would be to allocate first to a general-purpose reference semantic dialect and then lower to HiSEP-Q, to make things less error-prone.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
rainij: Funny thing is that I first learned about the word "hoist" in a context not involving loops or other control flow. Some languages _hoist_ function declarations to the top of a module. This way one can use them before they are defined. I good definition reads like this:

hoist = move to an earlier position that dominates the original
sink (dual) = move to a later position that post-dominates the original

This would cover my case and your case and give it a well defined meaning.

mlir/test/tools/qcc-opt/QCOToQVec/convert-qco-to-qvec.mlir
rainij: Because the test only cares about "static is not lowered". Currently this other pass relies on the fact that qubits come from static. Might change in the future but for now it is important hence a dedicated test.

mlir/include/qcc/Dialect/QVec/IR/QVec.h
julfarn:
```suggestion
/// Traces back the qubit at `index` in the vector `qubits` to a StaticOp and returns it if possible (null value if not).
```
is the convention we picked I believe

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
julfarn: Ah, these complaints were from before I read the implementation. Since you fiddle with the IR globally and with the program order, these tests actually make sense

mlir/test/tools/qcc-opt/Dialect/QVec/qvec-merge.mlir
julfarn: > This, like most passes, only look at the graph structure of the IR and don't care about order of operations.

I wrote this before reading the implementation, it's false. Hence test would have been fine, but good to remove if redundant.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
julfarn: I understand this, but I don't think it answers my question. If you attempt to trace back a value to a static op, but on the way you encounter an unknown op (leading to incomplete producers), you can't reach the static op.

The mechanism to trace back to static ops is via complete producers. With incomplete producers anywhere along the chain, the traceback will fail for some qubit. Hence if you can trace back all qubits to static ops, you also know that all producers are complete.

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
julfarn: Resolving as not super relevant

mlir/lib/Dialect/QVec/Transforms/QVecMerge.cpp
rainij: We only ever trace back to static if the producers are known (and even in that case it can of course fail). You probably assume trace back to static succeeds always, but this is not the case, we even filter out the things which have incomplete producers before tracing to static. Actually the trace back to static is a safety net on its own it is only needed because the producers can be incomplete. It is used to ensure the ops to be mergedoperate on disjoint qubits.

julfarn (approved):

PR #145, opened 2026-09-07
Fix cmake target check-qcc bug

rainij:
Running

```
cmake --preset my-dev
cmake --build build/my-dev --target check-qcc
```

fails on three tests. (One can avoid this by running `cmake --build build/my-dev` in between, which is the reason this did not surface in CI.) The reason for the failure is a missing `hisepq-elf2mem` dependency. This is fixed now.


EnriqueNB (approved):
Looks good.

PR #147, opened 2026-09-08
Build docs

rainij:
Added a cmake target `qcc-doc` (markdown) and `qcc-doc-html` (hugo) to build the documentation.


EnriqueNB (approved):
I could generate the .md and .html without problem. The implementation is as it should be. We could have a more fancy .html template to put the generated text but this is a problem of another PR.

PR #150, opened 2026-09-15
Complete QVec compilation pipeline.

### LINKED ISSUE #140: Use QVec dialect for HiSEPQ target
Replace the existing QIR based pipeline by a pipeline routed through the `qvec` dialect. Can be done in a two-step process (first establish the new route, then delete the old QIR based approach).

If possible already implement a qrisp integration test which targets hisepq. Factor out into a dedicated PR if the change would be too big.

# Ideas for implementation

Add target options to the target pass factory

```cpp
addLoweringPasses(PassManager&, const TargetOptions&)
```

This can be used to hold lower bound on `VLEN` and the qubit element width (QEW) . In the simplest case the info could be provided by an option:

```shell
# --min-vlen and friends only available for the hisepq target:
qcc --target=hisepq --min-vlen=256 --qubit-element-width=8 ...
```

The two existing passes `qvec-merge` and `convert-qvec-to-hisepq-intrinsics` can use this information already.

Remarks
- Similar info is often put into `-mcpu` options.
- Watch out for `llvm::RISCVISAInfo::getMinVLen()`.
- Watch out for ` llvm::MCSubtargetInfo` and `llvm::TargetSubtargetInfo`.
- Note that `min-vlen` is encoded as N in `+zvl<N>b`.

EnriqueNB:
The lowering pipeline to HiSEP through QIR has been substituted by the new pipeline through qco and qvec. The new machine options `minVLen` and  `QEW` have been added as command line options and as target options.


mlir/lib/Conversion/ToHiSEPQ/ConvertQVecToHiSEPQIntrinsics.cpp
```diff
@@ -285,6 +285,22 @@ struct MzOpLowering : public OpRewritePattern<MzOp> {
   HiSEPQMachine machine;
 };
 
+/// Erases an `aux` output recording op.
+///
+/// These ops ask for a classical result to be reported back to the host, which the QISA has no operation for, so the
+/// program ends up producing no output at all.
+///
+/// TODO: Lower them properly once HiSEP-Q specifies how a program reports its results. That is the same gap that makes
+/// `qvec.mz` lose its measurement result, so both should be fixed together.
```
julfarn, line 294: Help me understand: HiSEP-Q does not support the QIR concept of output recording. That's clear. But from an IR perspective, we don't really lose the information, right? If you measure, you get `i1` results in `qco`. With these you can do whatever you want. If you assume the HiSEP-Q processor to be a full processor, it could communicate these results in some way any other CPU handles IO. So the problem here is only: So far it's the user's responsibility to explicitly deal with the measurement results. We may want HiSEP-Q to follow the QIR "protocol" for output recording, and it hasn't been standardized yet how this should be done.

mlir/include/qcc/Target/HiSEPQ/HiSEPQTarget.h
```diff
@@ -19,11 +19,17 @@
 
 namespace qcc {
 
-/// `Target::addLoweringPasses` for the HiSEP-Q target.
-void addLoweringPassesHiSEPQ(mlir::PassManager& pm);
+/// `Target::addLoweringPasses` for the HiSEP-Q target, going through QIR.
+///
+/// TODO: Superseded by `addLoweringPassesHiSEPQViaQVec` and slated for removal.
+void addLoweringPassesHiSEPQViaQIR(mlir::PassManager& pm);
```
julfarn, line 25: If "slated for removal", can we not remove it now?

mlir/lib/Conversion/ToHiSEPQ/ConvertQVecToHiSEPQIntrinsics.cpp
EnriqueNB: What do you mean that from the IR perspective we don't loose the information? The problem is that right now we don't have an stablished way to report information back in HiSEPQ so I don't have the intrinsics to lower these instructions.

mlir/include/qcc/Target/HiSEPQ/HiSEPQTarget.h
EnriqueNB: So I had this conversation privately with @rainij. We could remove it because right now the HiSEQPViaQIR way is not used. We decided to leave it because we can always erase it and it could be useful as a template of an already tested way to lower to intrinsics through QIR.

mlir/include/qcc/Conversion/ToHiSEPQ/ToHiSEPQ.td
```diff
@@ -106,9 +106,14 @@ def ConvertQVecToHiSEPQIntrinsics : Pass<"convert-qvec-to-hisepq-intrinsics", "m
...
-      of `qvec.mz` is replaced with `poison`, i.e. the measurement result is lost.
-      TODO: Support this once HiSEP-Q specifies the corresponding QISA operation.
+      Nothing survives that would report a classical result back to the host, because the QISA has no operation that
+      does:
+
+      - `llvm.riscv.qv.mz` returns void and the QISA specifies no way to read a measurement back, so the classical
+        result of `qvec.mz` is replaced with `poison`, i.e. the measurement result is lost.
+      - `aux.record_int` and `aux.record_memref` are erased, so the program produces no output.
+
+      TODO: Support both once HiSEP-Q specifies the corresponding QISA operations.
```
rainij, line 116: IMO that is too verbose, the original message already contained the gist. I am OK with improving the original sentence, but we have to keep it brief (otherwise one gets tired reading docstrings and comments).

mlir/include/qcc/Target/HiSEPQ/HiSEPQTarget.h
rainij: Yes let it stay for now. We do not know what we are going to do about it. Could be that we take up development in the future for some reason.

mlir/include/qcc/Target/HiSEPQ/HiSEPQTarget.h
```diff
@@ -19,11 +19,17 @@
...
+
+/// `Target::addLoweringPasses` for the HiSEP-Q target, going through `qvec`.
+void addLoweringPassesHiSEPQViaQVec(mlir::PassManager& pm, const TargetOptions& targetOptions);
```
rainij, line 28: Should be renamed to `addLoweringPassesHiSEPQ` to emphasize that this is "the right one".

mlir/include/qcc/Target/TargetRegistry.h
```diff
@@ -33,17 +33,36 @@ struct NativeCodegenOptions {
 // entry carrying metadata plus a factory (`addLoweringPasses`) for the target's
 // behavior. If our implementation must be augmented follow LLVM's lead.
 
+/// Description of the machine a target lowers for, as selected on the command line.
+///
+/// The defaults describe the smallest machine we accept, so that a target which
+/// is handed default-constructed options still produces correct (if needlessly
+/// conservative) code.
```
rainij, line 40:
```suggestion
/// The defaults describe the smallest machine we accept.
```

Original was too verbose.

mlir/include/qcc/Target/TargetRegistry.h
```diff
@@ -33,17 +33,36 @@ struct NativeCodegenOptions {
...
+  std::function<void(mlir::PassManager&, const TargetOptions&)> addLoweringPasses;
   /// Emits native code for an already-lowered, LLVM-translated module. Null when
   /// the target has no native backend (e.g. QIR). Returns true on failure.
-  std::function<bool(llvm::Module&, llvm::raw_pwrite_stream&, const NativeCodegenOptions&)> emitNative;
+  std::function<bool(llvm::Module&, llvm::raw_pwrite_stream&, const NativeCodegenOptions&, const TargetOptions&)>
+      emitNative;
+  /// Whether `addLoweringPasses` reads the machine parameters in `TargetOptions`.
+  /// `qcc` rejects the corresponding flags for a target that does not, rather
+  /// than silently ignoring them.
+  bool usesMachineOptions = false;
```
rainij, line 65: I do not really understand why this `usesMachineOptions` is needed. Why do we care if a pass looks into it in a generic setting. I would prefer not having this option.

mlir/lib/Conversion/ToHiSEPQ/ConvertQVecToHiSEPQIntrinsics.cpp
rainij, line 294:
```suggestion
/// Erases output recording because HiSEP-Q does not support it.
///
/// TODO: Lower properly once HiSEP-Q specifies how a program reports its results.
```

Again to verbose.

mlir/lib/Conversion/ToHiSEPQ/EmitHiSEPQStart.cpp
```diff
@@ -21,8 +21,8 @@
 
 using namespace mlir;
 
-/// Whether `funcOp` carries the `entry_point` passthrough attribute.
-static bool isEntryPointFunc(LLVM::LLVMFuncOp funcOp) {
+/// Whether `funcOp` carries the `entry_point` passthrough attribute QIR marks its entry point with.
```
rainij, line 24:
```suggestion
/// Whether `funcOp` carries the `entry_point` passthrough attribute (for QIR)
```

This way it sounds more like an engineer wrote it. The original sounds more like coming from a poet :D .

mlir/lib/Conversion/ToHiSEPQ/EmitHiSEPQStart.cpp
```diff
@@ -34,6 +34,11 @@ static bool isEntryPointFunc(LLVM::LLVMFuncOp funcOp) {
   });
 }
 
+/// Whether `funcOp` is marked as an entry point.
+static bool isEntryPointFunc(LLVM::LLVMFuncOp funcOp) {
+  return funcOp->hasAttr(qcc::entryPointAttrName) || hasQIREntryPointPassthrough(funcOp);
+}
```
rainij, line 40: It is a bit strange that we check for either of the two. It looks a bit as if we do not know which one to expect. I would feel better if the function could decide which one it checks.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -15,8 +15,18 @@
...
 
+#include "qcc/Conversion/QCOToQVec/QCOToQVec.h"
+#include "qcc/Conversion/ToHiSEPQ/HiSEPQMachine.h"
 #include "qcc/Conversion/ToHiSEPQ/ToHiSEPQ.h"
+#include "qcc/Dialect/QVec/Transforms/Passes.h"
 #include "qcc/Target/QIR/QIRTarget.h"
+#include "qcc/Target/TargetRegistry.h"
+
+// The aggregate header, because `createConvertFuncToLLVMPass` is declared nowhere else.
+#include "mlir/Conversion/Passes.h"
```
rainij, line 26: I remember such comments on headers. I would remove it. Sounds like an excuse. It is just noise.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -35,13 +45,55 @@
...
-void addLoweringPassesHiSEPQ(mlir::PassManager& pm) {
+/// The cap `qvec-merge` gets so that it never builds an operation wider than the QV instructions can address.
+///
+/// Zero (unbounded) for invalid options; `convert-qvec-to-hisepq-intrinsics` reports those.
+static unsigned maxVectorizationFactor(const TargetOptions& targetOptions) {
+  using hisepq::HiSEPQMachine;
+  if (!HiSEPQMachine::isSupportedMinVLen(targetOptions.minVLen) ||
+      !HiSEPQMachine::isSupportedQubitElementWidth(targetOptions.qubitElementWidth)) {
+    return 0;
+  }
```
rainij, line 56: This sounds wrong to me: If stuff is not supported we just go unlimited?

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -35,13 +45,55 @@
 
 namespace qcc {
 
-void addLoweringPassesHiSEPQ(mlir::PassManager& pm) {
+/// The cap `qvec-merge` gets so that it never builds an operation wider than the QV instructions can address.
```
rainij, line 48: This is again this poet style comment. Can you rephrase it to something a normal person would write? I genuinely did not understand what it tried to tell me. I had to look into the implementation and the context (here the filename of the containing file).

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -35,13 +45,55 @@
...
+  pm.addPass(mlir::createConvertVectorToLLVMPass());
+  pm.addPass(mlir::createArithToLLVMConversionPass());
+  pm.addPass(mlir::createConvertControlFlowToLLVMPass());
+  pm.addPass(mlir::createConvertFuncToLLVMPass());
+
+  pm.addPass(qcc::createEmitHiSEPQStart());
+
+  // cleanup
+  pm.addPass(mlir::createCanonicalizerPass());
+  pm.addPass(mlir::createCSEPass());
```
rainij, line 92: Can you check this. I believe CSE does nothing if you run the Canonicalizer directly before it. The reason is that by default -canonicalize runs CSE iteratively (in fact _any_ pass using the greedy pattern matching does this by default).

mlir/lib/Conversion/ToHiSEPQ/ConvertQVecToHiSEPQIntrinsics.cpp
julfarn: I just want to be a bit cautious and not rely overly on this QIR design. With a risc-v core with quantum extensions as control unit, QIR's output recording is a communication protocol between a CPU and a CPU (with extensons, but that's not important here). I still have trouble to see why we need a custom new text-based protocol here. 

Anyways it's clear that lowering of output recording is not supported atm, so please resolve when read.

julfarn (approved):

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -50,7 +102,11 @@ bool emitNativeHiSEPQ(llvm::Module& module, llvm::raw_pwrite_stream& os, const N
   LLVMInitializeRISCVAsmPrinter();
   LLVMInitializeRISCVAsmParser();
 
-  const std::string attrsStr = "+experimental-xqv";
+  // `zvl<N>b` is how RISC-V spells a guaranteed lower bound on VLEN, so this is the same machine parameter the
+  // lowering above picked its vector types from; see `llvm::RISCVISAInfo::getMinVLen`. Handing it to the backend as
+  // well keeps the two from reasoning about different machines. Note that the other extensions imply a lower bound of
+  // their own, and the larger one wins.
```
rainij, line 108: To verbose. Please simplify it.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
```diff
@@ -60,12 +116,12 @@ bool emitNativeHiSEPQ(llvm::Module& module, llvm::raw_pwrite_stream& os, const N
     return true;
   }
 
-  llvm::TargetOptions targetOptions;
+  llvm::TargetOptions llvmTargetOptions;
```
rainij, line 119: Why the rename to `llvmTargetOptions`? If there is no good reason please revert.

mlir/test/tools/qcc-opt/ToHiSEPQ/emit-hisepq-start.mlir
```diff
@@ -21,3 +21,12 @@ llvm.func @helper() {
 // CHECK-DAG:     %[[ENTRY:.*]] = llvm.mlir.addressof @kernel
 // CHECK:         llvm.inline_asm has_side_effects{{.*}}"mv sp, $0{{.*}}jalr ra, 0($1){{.*}}", "r,r" %[[SP]], %[[ENTRY]]
 // CHECK:         llvm.unreachable
+
+// -----
+
+llvm.func @kernel() attributes { qcc.entry_point } {
```
rainij, line 27: What exactly does this test do? I see that the main difference is that it uses a different way to mark the entry point. Feels related to this function I commented on which doesn't know which one to check for. Feels wrong. We should decide for one. 

Context: The passthrough version is according to the QIR standard. Our qcc version is the one we use anywhere else. Since we avoid QIR anyway it might make sense to use the qcc variant throughout. But I did not think this through. Please check.

mlir/test/tools/qcc-opt/ToHiSEPQ/hisepq-bell.mlir
rainij, line 1: What is the replacement of this file?

mlir/test/tools/qcc/qrisp-integration/ghz.mlir
```diff
@@ -2,6 +2,9 @@
 // RUN: FileCheck %s --check-prefix=CHECK-QIR < %t.ll
 // RUN: qir-runner --file %t.ll -s 5 | FileCheck %s --check-prefix=CHECK-SIM
 
+// The same program taken all the way to HiSEP-Q QISA, in builds that have that target.
+// RUN: %if hisepq %{ qcc --target=hisepq --compile-to=native %s | FileCheck %s --check-prefix=CHECK-QISA %}
```
rainij, line 6: I would not add this to the qrisp-qir integration tests. This is super hidden in here. Also distracting for people only concerened with qrisp and QIR. We could open a new folder `qrisp-hisepq-integration` for better discoverability. We can still use the existing generator script to generate the mlir.

rainij: The existing folder could be renamed to `qrisp-qir-integration`.

mlir/test/tools/qcc/target-hisepq/bell-pairs.mlir
rainij, line 1: My feeling is that it would be nice to reformulate this in qrisp and add this to the folder `qrisp-hisepq-integration` mentioned elsewhere by me.

mlir/test/tools/qcc/target-hisepq/control-flow.mlir
```diff
@@ -1,5 +1,9 @@
 // RUN: qcc --target=hisepq --compile-to=native %s | FileCheck %s
 
+// XFAIL: *
+// `qc-to-qco` turns the `scf.if` below into a `qco.if`, which `convert-qco-to-qvec` has no pattern for, so the
+// pipeline stops there. Lowering it back to `scf.if` is a separate change; drop this XFAIL along with it.
+
```
rainij, line 6: Sounds like this should be turned into an issue. Please open and reference from here. (again comment is a bit verbose)

mlir/test/tools/qcc/target-hisepq/smoke.mlir
```diff
@@ -31,17 +31,22 @@ func.func @main() attributes { qcc.entry_point } {
...
 // CHECK:         qv.h     [[V1]], zero, 0
 // CHECK:         qv.cx    [[V1]], [[V2]], 0
 // CHECK:         qv.cx    [[V2]], [[V3]], 0
-// CHECK-DAG:     qv.mz    [[V1]], zero, 0
-// CHECK-DAG:     qv.mz    [[V2]], zero, 0
-// CHECK-DAG:     qv.mz    [[V3]], zero, 0
+// CHECK:         qv.mz    [[V1]], zero, 0
+
+// The measurements of qubits 1 and 2 are independent, so `qvec-merge` packs them into a single instruction over both
+// -- note the `vl` of 2. The QIR path emitted one instruction per qubit.
```
rainij, line 47: Please remove the comment. It doesn't matter what QIR did in the past. Comment just distracts.

mlir/test/tools/qcc/target-hisepq/vectorization-cap.mlir
```diff
@@ -0,0 +1,99 @@
+// RUN: qcc --target=hisepq --qubit-element-width=16 --compile-to=mlir %s | FileCheck %s
+
+// The vectorization factor `qvec-merge` is allowed to reach is the machine's capacity, so that it never builds an
+// operation wider than the QV instructions can address.
+//
+// At the default VLEN of 64 and a QEW of 16, the widest register group (LMUL 8, `vector<[32]xi16>`) addresses
+// `8 * 64/64 * 64/16 = 32` qubits. The forty gates below are all independent, so without a cap they would merge into
+// one operation on forty qubits, which no register group holds and the lowering would have to reject.
```
rainij, line 8: Far to verbose. A oneliner should be enough (if needed at all).

mlir/test/tools/qcc/target-hisepq/vectorization-cap.mlir
```diff
@@ -0,0 +1,99 @@
+// RUN: qcc --target=hisepq --qubit-element-width=16 --compile-to=mlir %s | FileCheck %s
```
rainij, line 1: should make min-vlen (or whatever the option is called) explicit.

mlir/include/qcc/Target/TargetRegistry.h
```diff
@@ -33,17 +33,36 @@ struct NativeCodegenOptions {
...
+/// is handed default-constructed options still produces correct (if needlessly
+/// conservative) code.
+///
+/// TODO: These are HiSEP-Q's parameters. Move them behind an `-mattr` string (`+zvl<N>b`).
+struct TargetOptions {
+  /// Guaranteed lower bound on VLEN, the vector register length in bits.
+  unsigned minVLen = 64;
+  /// QEW: how many bits one qubit index occupies in a qubit vector.
+  unsigned qubitElementWidth = 8;
+};
```
rainij, line 48: I am not sure if I feel good about setting defaults. In any case we should always set those values explicitly ourselves (e.g. in tests). Nothing should rely on those IMO. I tend to not having defaults, but I am unsure if there might be a true value (beyond giving an example that works).

mlir/tools/qcc/qcc.cpp
```diff
@@ -78,6 +78,13 @@ int main(int argc, char** argv) {
                                         cl::init("qir"), cl::value_desc("name"), cl::cat(qccCategory));
   const cl::opt<bool> listTargets("list-targets", cl::desc("List the available --target backends and exit"),
                                   cl::init(false), cl::cat(qccCategory));
+  const cl::opt<unsigned> minVLen("min-vlen",
+                                  cl::desc("Guaranteed lower bound on VLEN in bits; a power of two, at least 64 "
+                                           "(--target=hisepq only)"),
+                                  cl::init(qcc::TargetOptions{}.minVLen), cl::value_desc("bits"), cl::cat(qccCategory));
```
rainij, line 84: If we have defaults anywhere I could imagine here would be the right place (64 in that case). But again, I would not rely in tests on it. Similar for QEW.

rainij (changes requested):

mlir/include/qcc/Conversion/ToHiSEPQ/ToHiSEPQ.td
EnriqueNB: Changed.

mlir/include/qcc/Target/HiSEPQ/HiSEPQTarget.h
EnriqueNB: Changed.

mlir/include/qcc/Target/TargetRegistry.h
EnriqueNB: Changed.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: Removed.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: Changed.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: Refactored.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: This was renamed to not be confused from our newly created TargetOptions as this one comes from llvm::TargetOptions.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: Erased.

mlir/lib/Conversion/ToHiSEPQ/EmitHiSEPQStart.cpp
EnriqueNB: Changed.

mlir/lib/Conversion/ToHiSEPQ/ConvertQVecToHiSEPQIntrinsics.cpp
EnriqueNB: Changed.

mlir/include/qcc/Target/TargetRegistry.h
EnriqueNB: This is a workaround so that we have a way for the target to know if they use the MachineOptions. Otherwise, a possible input could be compiling to QIR and specifying the machine options. This is not correct. The correct way to solve this would be to define the -mattr setting and have the supported machine attributes of each machine be parsed againts each target. As this is a more involved approach, we stay with and create an issue to account for the needed change.

mlir/include/qcc/Target/TargetRegistry.h
EnriqueNB: Agreed, this is going to be resolved when we implement the per-target feature table that is associated with mattr. Right now this is a work around because it is the only machine we support.

mlir/lib/Conversion/ToHiSEPQ/EmitHiSEPQStart.cpp
EnriqueNB: This is the problem of maintaining the QIR route. This should be erased (or modified) when we decide what to do about that route.

mlir/test/tools/qcc-opt/ToHiSEPQ/emit-hisepq-start.mlir
EnriqueNB: Exactly, this check the new entry_point loweting. The fact that we also support the QIR route is a result of keeping that route working.

mlir/test/tools/qcc/target-hisepq/control-flow.mlir
EnriqueNB: Already addressed in a PoC PR and a related issue.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: Not silently: convert-qvec-to-hisepq-intrinsics rejects the same options right after, so the compile fails and the unbounded merge is never used. The 0 only exists because HiSEPQMachine can't be constructed from invalid values and the diagnostic lives in that pass. This is a workaround that is solved in the mattr PR.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
EnriqueNB: I checked, and it's actually not the case: canonicalize only deduplicate constants. So here it does work.

mlir/test/tools/qcc/target-hisepq/smoke.mlir
EnriqueNB: Removed.

mlir/test/tools/qcc/target-hisepq/vectorization-cap.mlir
EnriqueNB: Changed.

mlir/tools/qcc/qcc.cpp
EnriqueNB: To be changed in soon PR. Now a workaround.

mlir/test/tools/qcc/target-hisepq/vectorization-cap.mlir
EnriqueNB: Done.

mlir/test/tools/qcc/qrisp-integration/ghz.mlir
EnriqueNB: I would keep it target neutral as this is about Qrisp frontend.

EnriqueNB (commented):
@rainij Your comments have been addressed.

mlir/include/qcc/Target/TargetRegistry.h
rainij: See #154

mlir/include/qcc/Target/TargetRegistry.h
rainij: See #154

mlir/tools/qcc/qcc.cpp
rainij: See #154

mlir/lib/Conversion/ToHiSEPQ/EmitHiSEPQStart.cpp
rainij: I removed the QIR alternative.

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
rainij: Added a TODO that this is bad. To be resolved in #154 .

mlir/lib/Target/HiSEPQ/HiSEPQTarget.cpp
rainij: We checked together, CSE is off by default, so the code is right as is.

mlir/test/tools/qcc-opt/ToHiSEPQ/emit-hisepq-start.mlir
rainij: Removed the second one.

mlir/test/tools/qcc/target-hisepq/control-flow.mlir
rainij: See #155

mlir/test/tools/qcc/target-hisepq/bell-pairs.mlir
rainij: See #156.

rainij (approved):

